PR #5, opened 2025-07-10
Simple test for CNN

scoutcallens:
I've added a basic API test built for CNN for now that runs the CNN auth flow (and extraction, but nothing is shown on this end). To use it you will need to add a `CNN_USERNAME` and `CNN_PASSWORD` to your `.env` file.

To run the test manually, you'll need to spin up a server at `localhost:8000` (or any other local server, and just point the test to that instead). Once that is running you can run this command in the terminal:

`uv run pytest tests/api/test_auth_api.py -k "cnn" -v`

You should see the browser open to CNN and the username and password input will take place automatically.

In the future I want to add these tests to our container so they run on every PR. I also want to add an ACME suite of tests for an ACME local website that will test various methods of logging in. Step by step.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -3,5 +3,9 @@ ENVIRONMENT=local
 APP_NAME=getgather-local
 LOG_LEVEL=DEBUG
 
+# Authentication credentials
+CNN_USERNAME=
+CNN_PASSWORD=
+
 # Logging
 SENTRY_DSN=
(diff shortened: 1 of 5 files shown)


.env.template
```diff
@@ -3,5 +3,9 @@ ENVIRONMENT=local
 APP_NAME=getgather-local
 LOG_LEVEL=DEBUG
 
+# Authentication credentials
+CNN_USERNAME=
+CNN_PASSWORD=
```
ariya, line 8: This `.env` is typically for running the server and not for testing, so it's better to leave out env for tests (like these CNN credentials) as they can be supplied via other means (e.g. `export`).

PR #9, opened 2025-07-11
Add ACME Corp server and one test

scoutcallens:
This PR adds a server running locally at port 5001 for an ACME corp test website for programmatically testing the various login flows and capabilities of the getgather auth API. The idea is to have our own independent server (website) where we can establish any number of custom login flows to test the abilities of the auth orchestrator in a controlled environment.

The ACME server runs in a file called `acme_corp.py` and is basically a website with a list of endpoints taking you to the specific login flow pages. For example, I've added the `acme-email-password-checkbox` flow as the first test. ACME displays a link to this login flow. The actual page of the login flow is defined in the `acme_corp` folder as an extension to the localhost URL already in use (like `/auth/email-and-password-checkbox`). Then there is a corresponding yaml file for the email and password checkbox flow that defines the steps and possible pages necessary for logging in via this particular setup. It mirrors actual brand specs so that we can test various methods of login and potential pages seen during auth actions.

To run this and try it locally you'll have to do a few things:
1. Start the API server by running `uvicorn getgather.api.main:app --reload --host 127.0.0.1 --port 8000`
2. Start the ACME server by running `uvicorn tests.acme_corp.acme_corp:app --reload --host 127.0.0.1 --port 5001`
3. Run `uv run pytest tests/api/test_auth_api.py -k "acme-email-password-checkbox" -v`

The test should run at which point you'll see a window pop up with a browser going to the ACME site with the email and password and checkbox login style. If successful you'll see an output like this:

Screenshot 2025-07-11 at 10 58 59 AM

More additions to come:
* More ACME tests and pages
* Adding these to run on PRs so we know if new code has broken our ability to login to specific flows
* Adding an easy way to run both servers at once (probably via an sh script or something similar)
* CLI ability to run any brand as part of local development testing on new brands (as opposed to using a frontend to test)


pyproject.toml
```diff
@@ -19,6 +19,7 @@ dependencies = [
   # Browser Automation: Locked to specific versions to avoid breaking changes.
   "patchright==1.52.4",
   "requests>=2.32.4",
+  "python-fasthtml>=0.12.21",
```
ariya, line 22: This should be in `dev` of `dependency-groups`, as we need it only for ACME tests and not the main server.

ariya (approved):
Ship it 🚀

kpprasa (approved):
Nice transfer pr!

PR #11, opened 2025-07-14
Initial tests for the web UI

ariya:
Run the server first, e.g. via container:
```
docker build -t getgather .
docker run -p 8000:8000 --name getgather -d getgather
```

and then run the tests:
```bash
uv run pytest -s -m webui
```


scoutcallens:
> > Can't successfully get it to run. Seeing this error:
> 
> Perhaps try this first?
> 
> ```shell
> uv run playwright install chromium
> ```

`uv run` did the trick. Had tried several times before with just `playwright install` and had even removed it from the cache before adding it again. Worked after including `uv run`.

scoutcallens (approved):

PR #15, opened 2025-07-14
Add api tests to Github CI

scoutcallens:
Run the API authentication tests (right now only CNN and one ACME test) as part of every PR in the Github CI.

Also made some updates to the ACME test files by deprecating the linear tests and consolidating the state machine tests into the parent folder of brand_specs for tests. 

It's also important to note that for API tests, any brand yml defined in the "tests" folder is unavailable to the production image, so a local deployment must be spun up instead using uvicorn instead of the standard docker build and run.

Screenshot 2025-07-14 at 5 14 11 PM


.github/workflows/api-auth-tests.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - run: docker run -p 8000:8000 --name getgather -d getgather
+        timeout-minutes: 3
+
+      - name: Wait for the API server to start
+        run: while ! curl -s 'http://localhost:8000/health' | grep 'OK'; do sleep 1; done
+        timeout-minutes: 3
+
+      - name: Run the ACME Corp test server
+        run: uvicorn tests.acme_corp.acme_corp:app --reload --host 127.0.0.1 --port 5001 > >(tee /dev/stdout) 2> >(tee /dev/stderr
```
ariya, line 34: This needs `uv run -m uvicorn ...`.

PR #18, opened 2025-07-15
Custom tests for different style logins

scoutcallens:
Adds to the existing ACME Corp server several new login type tests based on real world scenarios. These are so we can accurately test our auth orchestrator's ability to follow different kinds of yamls pertaining to login scenarios that occur in the wild.

There are 12 new ACME-based tests in this PR, with more complicated ones on the way. These are simply ported over from an existing private repository where these have existed in the past. They will run on every PR to ensure changes do not affect our ability to successfully complete these login types.

Next step will be to port over the more complicated tests.

To run them, ensure both the API and ACME servers are running, then run `uv run pytest -n 7 -m "api" tests/api/test_auth_api.py`

To run an individual test simply add `-k "acme-email-then-otp" -v` to the previous command.

### DIFF
--- a/.github/workflows/api-auth-tests.yml
+++ b/.github/workflows/api-auth-tests.yml
@@ -45,7 +45,7 @@ jobs:
         timeout-minutes: 3
 
       - name: Run the API authentication tests
-        run: uv run pytest -n 2 -m "api" tests/api/test_auth_api.py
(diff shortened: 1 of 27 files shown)


ariya (approved):
Ship it 🚀

PR #22, opened 2025-07-21
Add two ACME tests w/ choices

scoutcallens:
Did some cleanup of the current ACME test structure as well as added in logic for two new tests that require choices on the part of the user (simulated here as a test).

One of the new tests allow for an email to be entered, at which point the user can either enter the password directly or login, OR choose OTP by clicking on a button, where they'll then have to enter the OTP and click log-in.

The other new test is the exact same but in reverse, so the user can enter OTP right away or select password and be redirected. 

This tests flows similar to Nike's, where it is unknown which method they will present first, but they always give the option for the other if the user wants it.


scoutcallens:
> Are there outdated tests/sets in `webui-auth-tests` again?

Shouldn't be

PR #29, opened 2025-07-28
feat: getgather MCP

faisalive:
Adds support for MCP (with HTTP Transport) to the existing codebase, allowing interaction with tools via the MCP Client.


ariya (approved):
Let's ship this 🚀

PR #53, opened 2025-08-06
use rich logger handler for better log format

bin-ario:


ariya:
@bin-ario why is there a lot of conflicts here?

ariya (approved):
LGTM

PR #59, opened 2025-08-07
Simplify host link and mcp session

bin-ario:
Rename `link` `session` to just `link` so it's not confused with browser session, http session or mcp session

Remove the use the mcp session_id since we can use link_id to keep track of auth

Replace mcp `SessionManager` with `BrandConnectionStore`, since we don't need to store connections per session, instead, we can just track all connected brands plainly (we will need to deal with expired or failed auth later)

Finally, simplify `BrowserProfile`, since it doesn't need to be lazily initialized anymore

Verify with unit tests and manual tests with Claude Desktop


getgather/mcp/main.py
```diff
@@ -5,45 +5,42 @@
...
+            return await call_next(context)
+
+        brand_id = BrandIdEnum(context.message.name.split("_")[0])
+        if BrandConnectionStore.is_brand_connected(brand_id):
+            return await call_next(context)
+
+        browser_profile_id = BrandConnectionStore.get_browser_profile_id(brand_id)
+        if not browser_profile_id:
+            browser_profile = BrowserProfile.create()
+            BrandConnectionStore.init_brand_state(brand_id, browser_profile.id)
```
faisalive, line 36: curious why do we need  different browser profile for each brand? Will there be a problem if we use same browser profile for all brands?

bin-ario: We originally designed this to isolate each user-brand pair. It seems easy to restore the connection. We can definitely revisit the decision at some point

PR #60, opened 2025-08-07
[draft] chore: boilerplates

broerjuang:
- [x] add shadcn + taiwindcss
- [x] add boilerplates to host different portal screens
- [ ]


broerjuang:
I’ll close this one since it’s based on feat/fe, which keeps changing with feedback. The other one uses main as the base, so it’s way easier to work with.

PR #63, opened 2025-08-07
MCP Documentation

scoutcallens:
Document the current tools available via MCP in a separate .md file on github. Added a link to the Getgather home page that redirects to said file.

### DIFF
--- a/getgather/mcp/tools.md
+++ b/getgather/mcp/tools.md
@@ -1,6 +1,8 @@
 # MCP Tools
 
-## A list of MCP tools available through the unified GetGather MCP Server
+## A sample list of MCP tools available through the unified GetGather MCP Server
+
+This list is non-exhaustive and only includes a sampling of the full set of MCP tools available on the server.
 
 ## Brand Specific Tools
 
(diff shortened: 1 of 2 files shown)


getgather/mcp/tools.md
```diff
@@ -0,0 +1,56 @@
...
+1. **Modular organization**: Each brand's functionality is in its own MCP server
+2. **Avoid naming conflicts**: Multiple brands might have similar tool names like get_orders
+3. **Clear namespacing**: You know which brand a tool belongs to by its prefix
+4. **Authentication context**: The middleware can extract the brand from the tool name by splitting the tool call.
+
+### Authentication Note
+
+Tools that require authentication will trigger the authentication process automatically through the MCP client if the user is not already authenticated for that brand.
+
+Here is a list of current brand specific tools:
```
ariya, line 28: Maintaining this list with the actual code will be hard. Perhaps just mention that this is the sample/preview of the tools and are not intended to be an exhaustive/complete list?

scoutcallens: Maybe later I can look into some sort of automated way to keep the list up to date with our MCP tools directory. It would be nice to always have a documented source of what's available. Although somewhat less important to know what goes into each tool since the protocol takes care of that by definition.

ariya (approved):

PR #65, opened 2025-08-07
Integrate browser-use for goodreads actions

bin-ario:
Use browser-use to accomplish goodreads actions for "adding to 'want to read'" and "get recommendations"

It sends browser-use agent progress back to mcp client if the client supports it

### DIFF
--- a/getgather/browser/profile.py
+++ b/getgather/browser/profile.py
@@ -48,7 +48,7 @@ class BrowserProfile(FreezableModel):
 
         return await browser_type.launch_persistent_context(
             user_data_dir=str(self.profile_dir(profile_id)),
-            headless=settings.HEADLESS,
+            headless=False,
             viewport=viewport_config,
             proxy=proxy,
         )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,9 @@ dependencies = [
   "fastmcp>=2.10.5",
   "requests>=2.32.4",
   "httpx>=0.28.1",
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -21,6 +21,9 @@ dependencies = [
   "fastmcp>=2.10.5",
   "requests>=2.32.4",
   "httpx>=0.28.1",
+  "browser-use>=0.5.9",
+  # Ping openai version for now since the latest version breaks browser-use
```
ariya, line 25: Ping -> Pin

getgather/browser/profile.py
```diff
@@ -48,7 +48,7 @@ async def launch(self, profile_id: str, browser_type: BrowserType):
 
         return await browser_type.launch_persistent_context(
             user_data_dir=str(self.profile_dir(profile_id)),
-            headless=settings.HEADLESS,
+            headless=False,
```
ariya, line 51: Intentionally hardcoded?

bin-ario: Do we want this to be configurable? I thought last time we decided to make it always not headless

ariya: I'm fine either way, but in that case better take out `HEADLESS` from settings and other related code in a different commit/PR altogether.

ariya (approved):
Ship it :rocket:

broerjuang:
Nice! 🚢

PR #78, opened 2025-08-08
feat: add replay functionality with rrweb player

faisalburhanudin:
## Replay

Add rrweb recording and replay frontend

https://github.com/user-attachments/assets/bae8e270-fb95-4272-bc3d-695baea539dc


frontend/src/components/rrweb-player.tsx
```diff
@@ -114,8 +115,8 @@ export function RRWebPlayer({ events }: RRWebPlayerProps) {
     const handleResize = () => {
       if (playerRef.current && containerRef.current) {
         // Recalculate dimensions on resize
-        const originalWidth = events[0]?.data?.width || 1920;
-        const originalHeight = events[0]?.data?.height || 1080;
+        const originalWidth = Number(events[0]?.data?.width) || 1920;
+        const originalHeight = Number(events[0]?.data?.height) || 1080;
```
broerjuang, line 119:
```ts
if (events && events.length) {
   const originalWidth = ...
}

```

frontend/src/components/rrweb-player.tsx
```diff
@@ -60,8 +61,8 @@ export function RRWebPlayer({ events }: RRWebPlayerProps) {
         if (isDestroyed || !containerRef.current) return;
 
         // Get original dimensions from first event
-        const originalWidth = events[0]?.data?.width || 1920;
-        const originalHeight = events[0]?.data?.height || 1080;
+        const originalWidth = Number(events[0]?.data?.width) || 1920;
```
broerjuang, line 64: please add guard to check first `events` exists and has more than one elements.

getgather/context.py
```diff
@@ -0,0 +1,37 @@
...
+    token = current_activity.set(activity)
+    try:
+        yield
+    finally:
+        # Reset the context and update end time
+        current_activity.reset(token)
+        Activity.update_end_time(
+            id=activity_id,
+            end_time=datetime.now(UTC),
+        )
```
broerjuang, line 37: just wonder why this needs to be on different file? I think it needs to be on Activity module . 

Activity:
  - add_activity
  - get_activity_by_id 
  - update_activity 
  
  
etc

package.json
```diff
@@ -27,6 +27,8 @@
     "react": "^19.1.0",
     "react-dom": "^19.1.0",
     "react-router": "^7.8.0",
+    "rrweb": "^2.0.0-alpha.4",
+    "rrweb-player": "^1.0.0-alpha.4",
```
broerjuang, line 31: Looking at this [link](https://github.com/rrweb-io/rrweb/blob/master/packages/rrweb-player/package.json#L3), the rrweb-player version can also be the same as rrweb itself. Why does it have a different version?

frontend/src/components/rrweb-player.tsx
```diff
@@ -0,0 +1,228 @@
...
+      className="rrweb-player"
+      style={{
+        width: "100%",
+        minHeight: "400px",
+        display: "flex",
+        justifyContent: "center",
+        alignItems: "center",
+      }}
+    />
+  );
```
faisalive, line 227: maybe you can use tailwind classname instead of inline style for concistency

`className="rrweb-player w-full min-h-[400px] flex justify-center items-center"`

frontend/src/components/rrweb-player.tsx
faisalburhanudin: updated

frontend/src/components/rrweb-player.tsx
faisalburhanudin: updated

getgather/context.py
faisalburhanudin: we dont have activity module right now

```
❯ fd activity
getgather/database/repositories/activity_repository.py
```

I think, I will just rename this module to activity.py

frontend/src/components/rrweb-player.tsx
faisalburhanudin: updated

getgather/auth_flow.py
```diff
@@ -58,63 +59,65 @@ async def auth_flow(
...
-                    browser_profile=browser_profile,
-                    nested_browser_session=True,
-                )
-                await extract_orchestrator.extract_flow()
-                extract_result = ExtractResult(
-                    profile_id=browser_profile.id,
-                    state=extract_orchestrator.state,
-                    bundles=extract_orchestrator.bundles,
+    # Use activity context manager for hosted link auth recording
+    async with track_activity(name="hosted_link", brand_id=str(brand_id)):
```
bin-ario, line 63: It's good to avoid multiple layers to indentation, which makes that code unreadable.
The easiest way is to create a wrapper function 
```
async def auth_flow(...):
  async with track_activity(...):
    await _auth_flow(...)  # rename the existing auth_flow to _auth_flow
```

getgather/database/repositories/rrweb_recordings_repository.py
```diff
@@ -0,0 +1,102 @@
...
+            return cls.model_validate(row)
+        return None
+
+    @classmethod
+    def create_recording(cls, activity_id: int, events: list[dict[str, Any]]) -> Self:
+        """Create a new recording from rrweb events."""
+        recording = cls.from_events(activity_id, events)
+        recording_id = cls.add(recording)
+        recording.id = recording_id
+        return recording
```
bin-ario, line 49: should these lines be done in `from_events`? this function seems redundant

and it's better for `DBModel.add` to return the entire object instead of just the `id`

getgather/database/repositories/rrweb_recordings_repository.py
```diff
@@ -0,0 +1,102 @@
...
+
+    @classmethod
+    def get_events_for_activity(cls, activity_id: int) -> list[dict[str, Any]] | None:
+        """Get the parsed events for an activity."""
+        if recording := cls.get_by_activity_id(activity_id):
+            return recording.events
+        return None
```
bin-ario, line 56: This classmethod is redundant and does not seem to be used

getgather/database/schema.sql
```diff
@@ -17,4 +17,19 @@ CREATE TABLE IF NOT EXISTS activities (
...
     created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
-);
\ No newline at end of file
+);
+
+-- rrweb recordings table
+CREATE TABLE IF NOT EXISTS rrweb_recordings (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    activity_id INTEGER NOT NULL UNIQUE,
+    events_json TEXT NOT NULL,
```
bin-ario, line 26: events_json could be a few hundreds megabytes, don't think you want to store them in the table. we are not using async db operation, loading lots of data from db is going to slow everything down

It's more efficient to store them as files.

getgather/database/repositories/activity_repository.py
```diff
@@ -14,6 +16,13 @@ class Activity(DBModel):
 
     table_name = "activities"
 
+    @classmethod
+    def get_all(cls) -> list[Self]:
+        """Get all activities ordered by start_time descending."""
+        query = f"SELECT * FROM {cls.table_name} ORDER BY start_time DESC"
+        rows = fetch_all(query)
+        return [cls.model_validate(row) for row in rows]
```
bin-ario, line 24: This potentially can cause performance issue if there are too many activities
We should add a `limit` to start with and add pagination later on if necessary

faisalburhanudin:
> There is a lot going on here, it would be good to separate the PR into a few smaller ones in the following order
> 
>     1. integrate rrweb to playwright, only store the recording as file
> 
>     2. frontend page to display a given recording file
> 
>     3. persistent storage, e.g., db and recording files

I think, I will just keeps the PR as it is, since the frontend part has already been reviewed. Adding 3 more PRs will make the reviewer need to review again, and I would have to reimplement something that I already did

getgather/database/schema.sql
faisalburhanudin: good idea, since the changes is big I will create ticket for now https://github.com/mcp-getgather/mcp-getgather/issues/119

getgather/database/repositories/rrweb_recordings_repository.py
faisalburhanudin: since the repository will be revamp (not using sqlite), I just resolve this issue for now

getgather/auth_flow.py
faisalburhanudin: done.

getgather/browser/session.py
```diff
@@ -65,6 +76,36 @@ async def start(self):
             self._context = await self.profile.launch(
                 profile_id=self.profile.id, browser_type=self.playwright.chromium
             )
+            await self._context.expose_function("saveEvent", self.save_event)  # type: ignore
+
+            # Add rrweb script loading to every page
+            await self._context.add_init_script("""
```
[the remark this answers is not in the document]
faisalburhanudin: It is our plan to add the toggle control in the settings, let me create ticket for it so I don't forget it https://github.com/mcp-getgather/mcp-getgather/issues/120

getgather/browser/session.py
```diff
@@ -24,6 +27,7 @@ def __init__(self, profile_id: str):
         self.profile: BrowserProfile = BrowserProfile(id=profile_id)
         self._playwright: Playwright | None = None
         self._context: BrowserContext | None = None
+        self._activity: Activity | None = None
```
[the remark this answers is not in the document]
faisalburhanudin: deleted

frontend/src/lib/api.ts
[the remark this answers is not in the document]
faisalburhanudin: Yes it is good, but we only have 2 endpoint right now, integrate with generator will be wasteful IMO. maybe after we have more than 5 endpoint that need to integrate then we can reconsider it

PR #81, opened 2025-08-08
Allow browser without login for public tools

kpprasa:
Previously if you wanted to search a bird on ebird without having authenticated there was just an error. This fixes that by allowing access to a browser when it's a public tool without needing the auth from the brand store. 

Also fixes a bug in one of the ebird mcp tools.

### DIFF
--- a/getgather/mcp/shared.py
+++ b/getgather/mcp/shared.py
@@ -106,10 +106,16 @@ async def extract(brand_id: BrandIdEnum) -> dict[str, Any]:
 
 async def start_browser_session(brand_id: BrandIdEnum) -> BrowserSession:
     """Start a browser session and return the page object."""
-    profile_id = BrandConnectionStore.get_browser_profile_id(brand_id)
-    if not profile_id:
-        raise ValueError(f"Profile ID not found for brand {brand_id}")
-    browser_profile = BrowserProfile(id=profile_id)
+    profile_id: str | None = None
+    if BrandConnectionStore.is_brand_connected(brand_id):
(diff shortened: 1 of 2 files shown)


scoutcallens (approved):

getgather/mcp/shared.py
```diff
@@ -106,10 +106,16 @@ async def extract(brand_id: BrandIdEnum) -> dict[str, Any]:
...
+    profile_id: str | None = None
+    if BrandConnectionStore.is_brand_connected(brand_id):
+        profile_id = BrandConnectionStore.get_browser_profile_id(brand_id)
+
+    if profile_id:
+        browser_profile = BrowserProfile(id=profile_id)
+    else:
+        # For public tools or unauthenticated brands, launch a fresh browser
+        # profile without persisting anything to the store.
+        browser_profile = BrowserProfile()
```
bin-ario, line 118: If you just need a new browser session, you can call 
```
from browser.session import browser_sessoin
async with browser_session(BrowserProfile()):
  ...
```
so this `start_browser_session` just handles the case where a consistent brand browser profile is required

getgather/mcp/brand/ebird.py
```diff
@@ -50,7 +50,18 @@ async def explore_species(
     """Explore species on Ebird from get_explore_species_list."""
     browser_session = await start_browser_session(brand_id=BrandIdEnum("ebird"))
```
bin-ario, line 51: it seems like this browser_session is not closed/stopped at the end

PR #82, opened 2025-08-08
MCP Inspector in the GetGather UI

scoutcallens:
This PR adds a new `/inspector` endpoint to the GetGather homepage (via a side panel button) that opens a running MCP inspector page in a new tab that is hooked up to our `/mcp` mount automatically. 

It works locally only (not sure if it would work in a published container) since it needs to run an npx command to access the MCP inspector's node. We redirect the browser to the Inspector’s UI with a token and the server URL prefilled, so it auto-connects to our MCP server at `/mcp`.

I tried it out and got it working on various tools in our MCP arsenal, and it was actually really cool. If you try it out make sure to wait a second after the clicking the button, as the inspector has to spin up first.

Disclaimer: I had GPT-5 write this code in cursor, so while it works (beautifully), I had to chat with it quite a bit afterward to understand what it did. I'm not a web expert so someone with a keener eye for this stuff should take a close look at the code and see how well it did.

### DIFF
--- a/getgather/api/frontend/old-index.html
+++ b/getgather/api/frontend/old-index.html
@@ -136,6 +136,7 @@
           type="text"
           placeholder="Search brand"
         />
+        <a href="/inspector" role="button" target="_blank" rel="noopener noreferrer">MCP Inspector</a>
       </aside>
       <main class="main-content">
         <div
--- a/getgather/api/main.py
+++ b/getgather/api/main.py
@@ -1,12 +1,16 @@
 import asyncio
+import os
+import secrets
 import socket
+import subprocess
 from contextlib import asynccontextmanager
 from datetime import datetime
 from os import path
 from typing import Final
+from urllib.parse import quote
 
 import httpx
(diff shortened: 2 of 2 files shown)


bin-ario:
to make npx work, the container needs to install nodejs
instead of an endpoint to start inspector, we can set the inspector to run in the dockerfile at http://0.0.0.0:6274, so the host can open access it via http://127.0.0.1:6274 similar to the live view

ariya:
Ideally the entire MCP inspector, including its UI in its fully glory, is available from `localhost:8000/inspector`.

If this merely launches `npx @modelcontextprotocol/inspector` and the user receives `localhost:6274` (but they potentially forgot to bind this port when launching the container), then the added value is very marginal, vis-a-vis launching it manually.

faisalburhanudin:
based on PR by @bin-ario I think we can close this PR
https://github.com/mcp-getgather/mcp-getgather/pull/153

PR #83, opened 2025-08-08
Add Hardcover spec and mcp

bin-ario:
Also make `flow_step` support `autoclick` field
Implement a `get_first_visible` helper function to handle cases where a selector could match single or multiple elements

Verify with Claude Desktop


ariya (approved):
Ship it :rocket:

PR #94, opened 2025-08-11
Add custom spec parse function

bin-ario:
Add the ability for a spec to use a custom parse function

Test it with Kindle extraction, where the book url doesn't appear in the html and needs to be custom computed


getgather/parse.py
```diff
@@ -143,24 +162,13 @@ async def parse_html(
...
-                                values.append(attr_value)
-                        else:
-                            text = await element.inner_text()
-                            if text:
-                                values.append(text)
-                    row[column.name] = values
+                    values = await asyncio.gather(*[
+                        _get_value(brand_id, column, element) for element in await elements.all()
+                    ])
+                    row[column.name] = [v for v in values if v is not None]
```
scoutcallens, line 168: Will not including None values throw off some indexing downstream? Looks like we were including Nones before, so just make sure this doesn't throw something off downstream

getgather/parse.py
```diff
@@ -78,26 +81,42 @@ async def parse(
...
         case "html":
-            return await parse_html(schema=schema, bundle_dir=bundle_dir, html_content=content)
+            return await parse_html(
+                brand_id=brand_id, schema=schema, bundle_dir=bundle_dir, html_content=content
+            )
         case _:
             raise ValueError(f"Format '{schema.format}' is not supported for parsing")
 
 
+async def _get_value(brand_id: BrandIdEnum, column: Column, element: Locator) -> str | None:
```
scoutcallens, line 108: Would it make sense to make brand_id optional so we don't have to pass it in all the time and update parse calls that don't need it (such as the ones in this PR)? Might be a reason you did it that I'm overlooking, thinking out loud here

bin-ario: the args here are not directly passed the the custom function so it doesn't really matter
we may need to change line 113 in the future where we actually pass args to the custom functions

getgather/parse.py
bin-ario: we were not including Nones, `values.append` was only called when the value is not None
if all values are None, you get an empty list, which is the behavior before

PR #97, opened 2025-08-11
add toggle for blocking unwanted resources (e.g. images)

kpprasa:
We may eventually want to block images etc at the proxy service level but first stab it doesn't seem to be as simple as I thought. 

This is a nice ratchet and we can eventually followup with that.

### DIFF
--- a/getgather/config.py
+++ b/getgather/config.py
@@ -25,6 +25,7 @@ class Settings(BaseSettings):
     BUNDLES_DIR: str = ""
     SCREENSHOTS_DIR: str = ""
     HEADLESS: bool = False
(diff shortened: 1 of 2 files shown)


getgather/config.py
```diff
@@ -25,6 +25,7 @@ class Settings(BaseSettings):
     BUNDLES_DIR: str = ""
     SCREENSHOTS_DIR: str = ""
     HEADLESS: bool = False
+    SHOULD_BLOCK_UNWANTED_RESOURCES: bool = False
```
bin-ario, line 28: I'd drop the `SHOULD_` prefix to make it shorter

bin-ario (approved):

PR #102, opened 2025-08-12
feat: integrate mcp brand settings

faisalive:
### MCP Server:
- Add middleware to filter MCP tools based on the GetGather Station settings page.
- Separate the MCP middleware into its own file.
- Add a new column `enabled` to `brand_states`. Also, make `browser_profile_id` nullable.
- Create a seeder for the `brand_states` table so on the first run, the app seeds the table based on the available MCP brands.
- Create new station endpoints: [GET] /station/brands and [PATCH] /station/brands/{brand_id} to toggle the enabled state.

### GetGather Station:
- Integrate the GetGather Station brand settings page with the API.

PS: Since we currently don’t have proper migrations, please delete your getgather.db after this PR is merged.

https://github.com/user-attachments/assets/ce08e617-d58b-4c62-b70f-0310d0ffd8f7

### DIFF
--- a/frontend/src/pages/Settings.tsx
+++ b/frontend/src/pages/Settings.tsx
@@ -6,7 +6,7 @@ import {
   CardHeader,
   CardTitle,
 } from "@/components/ui/card";
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import {
   Info,
   Download,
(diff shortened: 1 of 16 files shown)


frontend/src/pages/Settings.tsx
```diff
@@ -20,83 +20,70 @@ import {
...
+    setBrands(
+      brands.map((b) =>
+        b.brand_id === brand_id ? { ...b, enabled: !b.enabled } : b,
       ),
     );
+    apiService
+      .updateBrandEnabled(brand_id, !brand.enabled)
+      .then(() => {})
+      .catch((error) => {
+        console.error(error);
```
broerjuang, line 72: NIT: remove the log here. I think we already set this project with Sentry better to have sentry log this too. (Different issue)

frontend/src/pages/Settings.tsx
```diff
@@ -20,83 +20,70 @@ import {
...
+        );
+      });
   }
 
+  useEffect(() => {
+    apiService.fetchBrands().then((result) => {
+      setBrands(result);
+    });
+  }, []);
+
```
broerjuang, line 86: feels like we need to put this logic into its own hooks `useBrands` -> better if we can just fetch it once and store it globally.

frontend/src/pages/Settings.tsx
```diff
@@ -20,83 +20,70 @@ import {
...
-  },
-];
+  function handleError() {
+    if (src.endsWith(".svg")) {
+      setSrc(`/static/assets/logos/${brandId}.png`);
+      return;
+    }
+    if (src.endsWith(".png")) {
+      setSrc("/static/assets/logos/default.svg");
+    }
```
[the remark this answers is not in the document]
faisalive: this one is for fallback @broerjuang, since not all brands already have `.svg` file.

yuxicreate:
Thanks @ariya for the notes! There was also discoverability benefit of where people use one of the MCP may want to try out more, but it has the same challenge as #3 you listed. 
Agree that for the near term the focus should be more on the Activity Feeds first.

PR #107, opened 2025-08-12
Add ubereats spec and mcp

bin-ario:
Update Ubereats auth spec to handle different options

Add convenient testing code for extract and parse

### DIFF
--- a/getgather/parse.py
+++ b/getgather/parse.py
@@ -196,3 +196,17 @@ async def parse_html(
             parse_schema=schema,
             content=data,
         )
+
+
+# test with:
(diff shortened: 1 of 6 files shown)


getgather/parse.py
```diff
@@ -196,3 +196,17 @@ async def parse_html(
...
         )
+
+
+# test with:
+# python getgather/parse.py [brand] data/bundles/[profile_id]/[brand]
+if __name__ == "__main__":
+    import asyncio
+    import sys
+
+    from rich import print
```
ariya, line 207: Might as well promote this to the top level import!

PR #109, opened 2025-08-13
fix: Dockerfile, pinning docker base image to bookworm

faisalburhanudin:
## Summary
- Debian 13 "Trixie" was released Aug 9, 2025. Python 3.13-slim images started using Trixie by default, but Playwright doesn't support it yet and tries to install deprecated font packages.

## Before
image

## After
image

### DIFF
--- a/test
+++ b/test
@@ -1,0 +1,5 @@
+#!/bin/bash
+# Test file for CI testing
+echo "Running CI test..."
+echo "Test completed successfully"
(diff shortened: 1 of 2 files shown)


test
ariya, line 1: Is this file necessary?

faisalburhanudin: No, deleted

broerjuang (approved):

PR #116, opened 2025-08-14
Fix Nordstrom password sign in

bin-ario:
Update Nordstrom spec to make password sign in work, including wrong password case
Using message field to handle wrong password
Correctly handle navigation timeout

Note that SMS OTP is not fixed and requires a different solution

### DIFF
--- a/getgather/flow.py
+++ b/getgather/flow.py
@@ -2,7 +2,7 @@ import asyncio
 import json
 from typing import Any
 
-from patchright.async_api import Frame, Page
+from patchright.async_api import Frame, Page, TimeoutError
 
 from getgather.actions import (
     get_brand_function,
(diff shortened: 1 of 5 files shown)


ariya (approved):
LGTM

getgather/flow.py
```diff
@@ -130,16 +130,18 @@ async def _collect_needed_inputs(flow_state: FlowState, page: Page) -> StateProm
...
         prompts_list: list[InputPrompt] = []
         for fld in choice.fields_accept_input:
             prompts_list.append(await handle_field_prompt(page=page, field=fld))
         groups.append(
-            PromptGroup(
+            ChoicePrompt(
                 name=group_name,
+                prompt=choice.prompt,
+                groups=prompts_list,
                 prompts=prompts_list,
```
scoutcallens, line 145: I assume this "prompts" key is being removed after return reminder and others get caught up?
@bin-ario

bin-ario: prompts will be removed once this return-reminder PR is merged https://github.com/mcp-getgather/return-reminder/pull/14

PR #117, opened 2025-08-14
feat: sentry with proxy service

shabrina-ha:
solve: https://github.com/gather-engineering/dax/issues/866

### DIFF
--- a/getgather/startup.py
+++ b/getgather/startup.py
@@ -1,4 +1,6 @@
+import base64
 import logging
+from typing import Any
 
 import sentry_sdk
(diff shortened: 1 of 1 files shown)


scoutcallens (commented):
From my own exploration on this it looks correct, as long as you've tested it locally and can see the error message is gone.

I'll let Kiran have the final say though

kpprasa:
Apologies @shabrina-ha, I misled you by trying to solve the problem too fast. 

What you did should work, but is not the right long term solution I think. See this pr below. 

https://github.com/mcp-getgather/mcp-getgather/pull/121

PR #121, opened 2025-08-14
Change proxy env name

kpprasa:
# Motivation:
Sentry was throwing errors like: 
Screenshot 2025-08-14 at 11 20 42 AM

When running with a proxy. I initially thought the proper solution was to configure sentry to use the proxy which was done by @shabrina-ha in this PR: #117 

However, upon further reflection I realized that there should be no reason why Sentry's requests go through our proxy. In fact, we only setup our proxy config at the browser level and as such I became curious to find out why Sentry was routing through the proxy. I found out that the `HTTP_PROXY` env var is standard to the point of being assumed by Sentry's client and thus the problems were arising. By simply renaming the variable to `BROWSER_HTTP_PROXY`, Sentry doesn't know about it and the traffic flows as per normal (ie sentry traffic bypasses the proxy). 

Screenshot 2025-08-14 at 11 21 40 AM

Commensurate changes would need to be made in the env variables of any deployed instances.

### DIFF
--- a/getgather/browser/proxy.py
+++ b/getgather/browser/proxy.py
@@ -29,7 +29,9 @@ async def setup_proxy(
     """
     # Check if proxy service is configured
     if not settings.BROWSER_HTTP_PROXY or not settings.BROWSER_HTTP_PROXY_PASSWORD:
-        logger.info("No proxy configured (BROWSER_HTTP_PROXY and BROWSER_HTTP_PROXY_PASSWORD not set)")
+        logger.info(
+            "No proxy configured (BROWSER_HTTP_PROXY and BROWSER_HTTP_PROXY_PASSWORD not set)"
+        )
         return None
 
     # Use profile ID as base username


.env.template
```diff
@@ -9,5 +9,5 @@ SENTRY_DSN=
 OPENAI_API_KEY=
 BROWSER_USE_MODEL=
 # Proxy Settings (Optional)
-# HTTP_PROXY=http://your-proxy-server:port
-# HTTP_PROXY_PASSWORD=your-proxy-password
+# BROWSER_HTTP_PROXY=http://your-proxy-server:port
+# BROWSER_HTTP_PROXY_PASSWORD=your-proxy-password
```
scoutcallens, line 13: gnarly discovery

scoutcallens (approved):
I'll make the appropriate changes on the GCE containers running the proxy when this gets merged

PR #122, opened 2025-08-14
More robust text field input

bin-ario:
Added better support for handling multi input fields that auto-advances, such as Nordstrom OTP. It requires some specific keyboard events to trigger submission

Cleaned up input types
- removed `email`, which is just an alias of `text`
- added `text_multi` and `text_multi_auto` to distinguish different types of multi input fields
- `selectors` is redundant since 'text_multi` can indicate multiple matches of `selector`

Nordstrom otp page also needs improvements on page detection scoring function so that we can correctly detect otp form

### DIFF
--- a/getgather/actions.py
+++ b/getgather/actions.py
@@ -140,29 +140,31 @@ async def handle_fill_single(frame: Frame | Page, field: Field, value: str) -> N
 
 
 async def handle_fill_multi(frame: Frame | Page, field: Field, values: str | list[str]) -> None:
-    """Fill repeated elements selected by `field.selectors`."""
+    """Fill repeated elements for text_multi type fields."""
     await frame.wait_for_timeout(500)
     # Coerce single string into a list for uniform processing.
     vals: list[str] = list(values) if isinstance(values, str) else values
-    outer = frame.locator(field.selectors)  # type: ignore[arg-type]
+    outer = frame.locator(field.selector)  # type: ignore[arg-type]
     await outer.first.wait_for(state="attached", timeout=5_000)
 
-    inner = frame.locator(f"{field.selectors} >>> input")
+    inner = frame.locator(f"{field.selector} >>> input")
     if await inner.count() > 0:
         await inner.first.wait_for(state="attached", timeout=5_000)
         target = inner
     else:
         # Fallback if no match
-        await frame.locator(f"css={field.selectors}").first.wait_for(
-            state="attached", timeout=5_000
-        )
-        target = frame.locator(f"css={field.selectors}")
+        await frame.locator(f"css={field.selector}").first.wait_for(state="attached", timeout=5_000)
+        target = frame.locator(f"css={field.selector}")
 
-    for idx, val in enumerate(vals):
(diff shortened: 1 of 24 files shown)


getgather/actions.py
```diff
@@ -140,29 +140,31 @@ async def handle_fill_single(frame: Frame | Page, field: Field, value: str) -> N
...
+
+    if field.type == "text_multi_auto":
+        await target.first.click()
+        await target.first.press_sequentially("".join(vals), delay=100)
+    else:
+        for idx, val in enumerate(vals):
+            logger.info(f"📝 Filling {field.name} {idx + 1}/{len(vals)}")
+            await target.nth(idx).clear()
+            await target.nth(idx).press_sequentially(val, delay=100)
+            await frame.wait_for_timeout(200)
```
scoutcallens, line 167: I feel like we don't really need to update this because even if the cursor auto-advances, it doesn't hurt to clear and press sequentially on every input. Might be overkill.

getgather/connectors/brand_specs/americanairlines/specs.yml
```diff
@@ -27,8 +27,8 @@ auth:
       type: wait
       selector: hp-account-dropdown-button#accountButtonDesktop
     - name: mfa_code
-      type: text
-      selectors: '[type="text"][inputmode="numeric"]'
+      type: text_multi
+      selector: '[type="text"][inputmode="numeric"]'
```
scoutcallens, line 31: Smart change, there was no real reason to have both `selectors` and `selector`

getgather/connectors/brand_specs/nordstrom/specs.yml
```diff
@@ -27,10 +27,13 @@ auth:
       selector: label:has-text("Keep me signed in.")
       expect_nav: false
     - name: otp
-      type: text
+      score: 10
```
scoutcallens, line 30: Why do we need to make the score 10 here? the logic should identify this page just fine without this

ariya:
> * removed `email`, which is just an alias of `text`

This might be obsoleted later with the middle-man's distillation, but technically if the field is necessary for the hosted link, `input[type=email]` is still useful for the web browser, vis-a-vis a regular input, since it allows the browser to autofill it (e.g. using its built-in password manager) as well doing some validation.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/email

getgather/actions.py
bin-ario: The previous approach doesn't work for Nordstrom because the page auto advances the input boxes and doesn't submit at the end. It probably wants a specific sequence of keyboard events

getgather/connectors/brand_specs/nordstrom/specs.yml
bin-ario: to make the otp page scores higher than the verification choice page

bin-ario:
> but technically if the field is necessary for the hosted link, `input[type=email]` is still useful for the web browser, vis-a-vis a regular input, since it allows the browser to autofill it (e.g. using its built-in password manager) as well doing some validation.

good point, i didn't notice we actually use `type` in our frontend display

PR #128, opened 2025-08-15
feat: add search purchase history

broerjuang:
add amazon search purchase history

### DIFF
--- a/getgather/mcp/brand/amazon.py
+++ b/getgather/mcp/brand/amazon.py
@@ -100,3 +100,44 @@ async def get_browsing_history() -> dict[str, Any]:
     await page.wait_for_selector("div[class*='desktop-grid']")
     html = await page.locator("div[class*='desktop-grid']").inner_html()
     return {"browsing_history_html": html}
+
+
+@amazon_mcp.tool(tags={"private"})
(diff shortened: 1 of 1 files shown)


getgather/mcp/brand/amazon.py
```diff
@@ -100,3 +100,44 @@ async def get_browsing_history() -> dict[str, Any]:
...
+) -> dict[str, Any]:
+    """Search purchase history of a amazon."""
+    browser_session = await start_browser_session(brand_id=BrandIdEnum("amazon"))
+    page = await browser_session.page()
+    await page.goto(
+        f"https://www.amazon.com/your-orders/search/ref=ppx_yo2ov_dt_b_search?opt=ab&search={keyword}"
+    )
+    await page.wait_for_selector("div.a-section.a-spacing-none.a-padding-small")
+    await page.wait_for_timeout(1000)
+    html = await page.locator("div.a-section.a-spacing-none.a-padding-small").inner_html()
```
bin-ario, line 117: why not make this an `extract` spec in spec yaml?
we should be able to support multiple bundles in extract, and client can specify which bundle it wants to fetch

PR #144, opened 2025-08-19
fix: search product on amazon

shabrina-ha:
Background 

The parse function took too long when we give the html directly, because it tries to create a new headless instance based on the html given. When using the parse, we already have the page and can use that to find what we need. 

This PR is part of #129. We're solving too many problems with undecided solution in that PR, so I'm creating a new PR to tackle each one to make it clearer. 

Can search product with no problem:
CleanShot 2025-08-20 at 12 43 53@2x

### DIFF
--- a/getgather/mcp/brand/amazon.py
+++ b/getgather/mcp/brand/amazon.py
@@ -51,6 +51,7 @@ async def search_product(
                 {"name": "currency", "selector": "span.a-price-symbol"},
                 {"name": "rating", "selector": "span.a-icon-alt"},
                 {"name": "image_url", "selector": "img.s-image", "attribute": "src"},
+                {"name": "reviews", "selector": "div[data-cy='reviews-block']"},
             ],
         })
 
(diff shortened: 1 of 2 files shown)


getgather/mcp/brand/amazon.py
```diff
@@ -30,28 +30,35 @@ async def search_product(
...
+                    "selector": "div[data-cy='title-recipe'] a",
+                    "attribute": "href",
+                },
+                {"name": "price", "selector": "span.a-price-whole"},
+                {"name": "price_fraction", "selector": "span.a-price-fraction"},
+                {"name": "currency", "selector": "span.a-price-symbol"},
+                {"name": "rating", "selector": "span.a-icon-alt"},
+                {"name": "image_url", "selector": "img.s-image", "attribute": "src"},
+            ],
+        })
```
broerjuang, line 55: could you please add `reviews` here?

shabrina-ha: PR description updated, as shown, reviews is added

broerjuang (approved):
I think in the long run we can remove this else branch. Since we’re already inside the page, we don’t need to open another browser just for parsing. But let’s leave that cleanup for another PR.

PR #154, opened 2025-08-20
Add mcp docs page

bin-ario:
Screenshot 2025-08-20 at 4 51 35 PM

### DIFF
--- a/getgather/api/main.py
+++ b/getgather/api/main.py
@@ -22,11 +22,11 @@ from getgather.config import settings
 from getgather.database.migrate import run_migration
 from getgather.hosted_link_manager import HostedLinkManager
 from getgather.logs import logger
-from getgather.mcp.main import create_mcp_apps
+from getgather.mcp.main import MCPDoc, create_mcp_apps, mcp_app_docs
+from getgather.startup import startup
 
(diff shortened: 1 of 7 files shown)


broerjuang (approved):

getgather/api/main.py
```diff
@@ -232,10 +232,68 @@ async def extended_health():
...
 app.include_router(activities_router)
 app.include_router(brands_router)
 app.include_router(auth_router)
 app.include_router(link_router)
-for bundle_name, mcp_app in mcp_apps.items():
-    route_name = "/mcp" if bundle_name == "all" else f"/mcp-{bundle_name}"
-    app.mount(route_name, mcp_app)
+for mcp_app in mcp_apps:
+    app.mount(mcp_app.route, mcp_app.app)
+
```
broerjuang, line 295: on this we also needs to add routes for frontend to access `index.html` similar like `activities` routes 

```
app.get("/mcp-docs")
def activities():
    file_path = path.join(path.dirname(__file__), "frontend", "index.html")
    with open(file_path) as f:
        template = Template(f.read())
    rendered = template.render()
    return HTMLResponse(content=rendered)

```

This will make sure that server knows what to render when user hit the routes directly.

PR #155, opened 2025-08-20
Fix: simplify RRWeb recording for whole page

faisalburhanudin:
Simplified RRWeb injection script to record whole page instead of individual iframes. Updated configuration to use standard rrweb options with canvas and font recording enabled.


broerjuang (approved):
Looks good to me 🚢 it

PR #176, opened 2025-08-27
Manual distillation test via CLI

ariya:
This is mainly copied from Middleman with very little modification. I plan to have a follow-up which will do the following:

* Use a proper and richer logging
* Avoid custom browser code and reuse the existing abstraction
* Switch to idiomatic Pydantic

For now, try this by running:

```bash
./getgather.py run goodreads.com/review
```

and you'll be prompted for email and password. If nothing goes wrong, the list of books will be printed out in its original HTML table.

And to run a simple distillation:

```bash
./getgather.py distill https://www.goodreads.com/user/sign_in
```

### DIFF
--- a/getgather.py
+++ b/getgather.py
@@ -1,0 +1,322 @@
+#!/usr/bin/env python3
+
+import argparse
+import asyncio
+import os
+import sys
+import urllib.parse
+from typing import TypedDict, cast
+
+import nanoid
+import pwinput
+from bs4 import BeautifulSoup
+from bs4.element import Tag
+from patchright.async_api import BrowserContext, Page, async_playwright
+
+from getgather.distill import Pattern, distill
+
+NORMAL = "\033[0m"
+BOLD = "\033[1m"
+YELLOW = "\033[93m"
+MAGENTA = "\033[35m"
+RED = "\033[91m"
+GREEN = "\033[92m"
+CYAN = "\033[36m"
+
+ARROW = "⇢"
+CROSS = "✘"
+
+FRIENDLY_CHARS = "23456789abcdefghijkmnpqrstuvwxyz"
+
+
+async def sleep(seconds: float):
+    await asyncio.sleep(seconds)
+
+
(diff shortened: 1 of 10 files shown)


getgather.py
```diff
@@ -0,0 +1,322 @@
...
+async def terminate(page: Page, distilled: str) -> bool:
+    document = parse(distilled)
+    stops = document.find_all(attrs={"gg-stop": True})
+    if len(stops) > 0:
+        print("Found stop elements, terminating session...")
+        return True
+    return False
+
+
+playwright_instance = None
```
broerjuang, line 166: Should we name it `patchright_instance` so it matches what we use?

getgather.py
```diff
@@ -0,0 +1,322 @@
...
+        directory, headless=False, viewport={"width": 1920, "height": 1080}
+    )
+
+    page = context.pages[0] if context.pages else await context.new_page()
+    return {"id": id, "hostname": hostname, "location": location, "context": context, "page": page}
+
+
+def load_patterns() -> list[Pattern]:
+    patterns: list[Pattern] = []
+    for name in [f for f in search("./getgather/connectors/brand_specs") if f.endswith(".html")]:
```
broerjuang, line 108: can we use [glob](https://docs.python.org/3/library/glob.html) instead custom search function? 

```py
def load_patterns() -> list[Pattern]:
    patterns: list[Pattern] = []
    for name in glob.glob("./getgather/connectors/brand_specs/**/*.html", recursive=True):
        with open(name, "r", encoding="utf-8") as f:
            content = f.read()
        patterns.append({"name": name, "pattern": parse(content)})
    return patterns
```

broerjuang (approved):
LGTM, tiny suggestion.

getgather.py
ariya: Great point, thanks @broerjuang!
I'll do that as a follow up, and for Middleman as well.

PR #181, opened 2025-08-29
feat: add iframe support for distillation

faisalive:
Handle elements that live inside an iframe with `gg-frame` as the iframe selector


faisalive:
> Looks great @faisalive! Can you add this to [middleman](https://github.com/mcp-getgather/middleman) first?

Sure, will add to middleman

PR #182, opened 2025-08-29
feat: embed live view in station

faisalive:
embed live view inside station web (iframe)

Screenshot 2025-08-29 at 19 25 46

### DIFF
--- a/frontend/src/pages/LiveView.tsx
+++ b/frontend/src/pages/LiveView.tsx
@@ -1,31 +1,18 @@
-import { Button } from "@/components/ui/button";
-import EmptyState from "@/components/EmptyState";
 import PageHeader from "@/components/PageHeader";
-import { Activity, Play } from "lucide-react";
 
 export default function LiveView() {
   return (
-    <div className="max-w-6xl mx-auto px-6 py-8">
+    <div className="max-w-6xl mx-auto px-6 h-full flex flex-col min-h-0 overflow-hidden">
       <PageHeader
         title="Live View"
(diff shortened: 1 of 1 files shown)


bin-ario (approved):

frontend/src/pages/LiveView.tsx
```diff
@@ -1,31 +1,18 @@
...
-        description="Start a getgather operation to see real-time monitoring here."
-        action={
-          <Button className="bg-indigo-600 hover:bg-indigo-700">
-            <Play className="w-4 h-4 mr-2" />
-            Start Monitoring
-          </Button>
-        }
       />
+      <div className="flex-1 min-h-0 overflow-hidden">
+        <iframe
```
ariya, line 11: Is there a way to maintain the aspect ration suitable for 1920x1080? That way, it will fit the desktop view without any filler gap.

frontend/src/pages/LiveView.tsx
ariya, line 1: As a follow-up, there should be some run-time detection to show/hide this live view, since it makes sense only when running inside a container.

A possible solution, `ENABLE_LIVE_VIEW` env which is default to false, but set to true in the Dockerfile.

frontend/src/pages/LiveView.tsx
faisalive: Sure, here's the result

https://github.com/user-attachments/assets/4b42dacb-2c97-4c58-bb6c-a70d67157ecc

PR #187, opened 2025-09-02
experiment: use stagehand

broerjuang:
Tried running Stagehand with our project. It uses Playwright under the hood, and we’re stuck at the auth part

https://github.com/user-attachments/assets/52a26172-ffc0-4402-b0a9-05fabe001d6c


broerjuang:
it's working on #194

PR #190, opened 2025-09-02
disable live view in multi user mode

bin-ario:
when auth is enabled, the service is running in multi user mode in cloud, so we should disable live view for privacy

made frontend page config dynamic based on the response of a new api `/api/station`

### DIFF
--- a/getgather/api/api.py
+++ b/getgather/api/api.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from fastapi import FastAPI
 from fastapi.responses import PlainTextResponse
 from fastapi.routing import APIRoute
+from pydantic import BaseModel
 
 from getgather.api.routes.activities.endpoints import router as activities_router
 from getgather.api.routes.auth.endpoints import router as auth_router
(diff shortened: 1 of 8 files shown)


getgather/api/api.py
```diff
@@ -35,6 +36,29 @@ def custom_generate_unique_id(route: APIRoute) -> str:
...
+        StationPage(path="", label="Get Started", name="GetStarted"),
+        StationPage(path="activities", label="Activities", name="Activities"),
+        StationPage(path="replay", label="Replay", name="Replay"),
+        StationPage(path="mcp-docs", label="MCP Docs", name="McpDocs"),
+    ]
+    if not settings.multi_user_enabled:
+        pages.append(StationPage(path="live-view", label="Live View", name="LiveView"))
+    return StationConfig(pages=pages)
+
+
```
broerjuang, line 61: I think since the goal is just to enable/disable a page or pass config to the frontend, we don’t really need the server to tell the frontend what to render. Something like [vite-plugin-runtime-env](https://github.com/micha149/vite-plugin-runtime-env) should be enough. wdyt?

ariya: @broerjuang Will this work with the production bundle? We don't run Vite in the container, so I'm curious what the mechanics will be to get the env variable.

bin-ario:
going to use a build env variable approach https://github.com/mcp-getgather/mcp-getgather/pull/196

PR #194, opened 2025-09-03
feat: stagehand agent

broerjuang:
This PR adds the Stagehand implementation where the handoff between our session and Stagehand is done by mounting the same profile. By default, Stagehand runs on Playwright.

Screenshot 2025-09-04 at 10 58 19

**Metrics**
TODO

### DIFF
--- a/getgather/mcp/stagehand_agent.py
+++ b/getgather/mcp/stagehand_agent.py
@@ -1,13 +1,16 @@
-"""
-Stagehand agent implementation for AI-powered browser automation.
-
-This module provides a functional interface for creating and using Stagehand agents
-with the existing getgather browser profile system for session persistence.
-"""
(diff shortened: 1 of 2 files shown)


getgather/mcp/stagehand_agent.py
```diff
@@ -0,0 +1,107 @@
...
+            "--disable-dev-shm-usage",
+            "--disable-blink-features=AutomationControlled",
+            "--disable-extensions",
+        ],
+    }
+
+    # Add userDataDir for session persistence if available
+    if user_data_dir:
+        launch_options["userDataDir"] = user_data_dir
+
```
broerjuang, line 66: we can verify on this line that `user_data_dir` is found

broerjuang: my bad, `launch_options` should be `launch_options["user_data_dir"]`

PR #196, opened 2025-09-03
use MULTI_USER_ENABLED to toggle live view

bin-ario:


frontend/src/main.tsx
```diff
@@ -18,6 +18,9 @@ import McpDocs from "./pages/MCPDocs";
 import NotFound from "./pages/NotFound";
 import { ReplayPage } from "./pages/Replay";
 import Start from "./pages/Start";
+import { liveViewEnabled } from "./pages/config";
+
+console.debug(`MULTI_USER_ENABLED: ${import.meta.env.MULTI_USER_ENABLED}`);
```
broerjuang, line 23: NIT: I think we should remove this.

vite.config.ts
```diff
@@ -1,29 +1,35 @@
...
+  return {
+    root: path.resolve(__dirname, "frontend"),
+    plugins: [react(), tailwindcss()],
+    esbuild: { pure: mode === "production" ? ["console.debug"] : [] }, // Remove console.debug in production
+    build: {
+      outDir: path.resolve(__dirname, "getgather", "frontend"),
+      assetsDir: "__assets",
+    },
+    define: {
+      "import.meta.env.MULTI_USER_ENABLED": env.MULTI_USER_ENABLED === "true",
```
broerjuang, line 17: This value is still set at build time, but I think that’s fine for now since we’re providing a default anyway.

Dockerfile
```diff
@@ -1,6 +1,9 @@
 # Stage 1: Frontend Builder
 FROM mirror.gcr.io/library/node:22-alpine AS frontend-builder
 
+ARG MULTI_USER_ENABLED=false
+ENV MULTI_USER_ENABLED=${MULTI_USER_ENABLED}
```
broerjuang, line 5: If we plan to expose this value to the frontend later, maybe we should add a prefix so it’s easier to tell which ones are public vs private. In Vite they use the VITE_ prefix.

vite.config.ts
bin-ario: It's intended to be used at build time

Dockerfile
bin-ario: MULTI_USER_ENABLED is shared between backend and frontend, trying to use a single variable so we don't end up misconfigured

broerjuang: Got it. Prefixing is really just a convention to make it obvious which ones are safe to expose. If we’re confident about managing that internally, keeping MULTI_USER_ENABLED as-is should be fine. Otherwise, adding something like PUBLIC_ or VITE_ later could help us avoid accidental leaks.

PR #200, opened 2025-09-04
Refactor the distillation by moving functions to the shared logic

ariya:
This is important for the upcoming distillation-based MCP tools.

To verify, as usual just run it with the example, e.g.:

```
./getgather.py run https://goodreads.com/review
```

and then supply the credentials until the loop terminates with the HTML version of the bookshelf.

### DIFF
--- a/getgather.py
+++ b/getgather.py
@@ -6,146 +6,25 @@ import os
 import sys
 import urllib.parse
 from glob import glob
-from typing import cast
-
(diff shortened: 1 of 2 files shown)


broerjuang:
Screenshot 2025-09-04 at 09 29 54

try it and it works great.

broerjuang (approved):

PR #203, opened 2025-09-04
disable more features in multi-user mode

bin-ario:
disable activities, inspector, and replay in multi-user mode


ariya (approved):

ariya:
I don't have any strong preference one way or another, but if the MCP inspector is only usable in the single-user mode, we may just let people use the stand-alone/external inspector and take out the built-in inspector at some point.

PR #215, opened 2025-09-05
use fastmcp pypi

bin-ario:
use the new release 2.12.2

### DIFF
--- a/getgather/mcp/auth.py
+++ b/getgather/mcp/auth.py
@@ -68,7 +68,7 @@ def setup_mcp_auth(app: FastAPI, mcp_routes: list[str]):
     auth_middleware = [
         Middleware(
             RequireAuthMiddlewareCustom,  # verify auth for MCP routes
-            getattr(github_auth_provider, "required_scopes", None) or [],
+            github_auth_provider.required_scopes,
         ),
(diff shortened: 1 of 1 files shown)


getgather/mcp/auth.py
```diff
@@ -69,7 +69,6 @@ def setup_mcp_auth(app: FastAPI, mcp_routes: list[str]):
         Middleware(
             RequireAuthMiddlewareCustom,  # verify auth for MCP routes
             getattr(github_auth_provider, "required_scopes", None) or [],
-            github_auth_provider.get_resource_metadata_url(),
```
bin-ario, line 72: `get_resource_metadata_url` no long exists and we don't seem to need it

getgather/mcp/auth.py
```diff
@@ -68,8 +68,7 @@ def setup_mcp_auth(app: FastAPI, mcp_routes: list[str]):
     auth_middleware = [
         Middleware(
             RequireAuthMiddlewareCustom,  # verify auth for MCP routes
-            getattr(github_auth_provider, "required_scopes", None) or [],
-            github_auth_provider.get_resource_metadata_url(),
```
bin-ario, line 72: get_resource_metadata_url no long exists and we don't seem to need it

PR #230, opened 2025-09-08
Change "auth" to "signin" for brand yamls

scoutcallens:
All the brand yamls (product and tests) have "auth" as the delineator for the flow state machine during sign in to the various brand websites. This PR changes that to be "signin" instead.


ariya (approved):

broerjuang (approved):
🚢 it

PR #231, opened 2025-09-09
add mcp client test nytimes

faisalive:
add test to make sure distill function (with JSON parse) work properly

### DIFF
--- a/tests/mcp/test_nytimes.py
+++ b/tests/mcp/test_nytimes.py
@@ -1,4 +1,4 @@
-"""Tests for BBC Tools: login, get_bookmarks."""
+"""Tests for NYTimes Tools: get_bestsellers_list."""
 
 import json
 import os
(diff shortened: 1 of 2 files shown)


broerjuang (approved):
nice

tests/mcp/test_nytimes.py
```diff
@@ -0,0 +1,26 @@
+"""Tests for BBC Tools: login, get_bookmarks."""
+
+import json
+import os
+
+import pytest
+from fastmcp import Client
+from mcp.types import TextContent
+
+config = {"mcpServers": {"getgather": {"url": f"{os.environ.get('HOST')}/mcp"}}}
```
ariya, line 10: what happens if `HOST` not defined? should there be a fallback default (useful for CI run)?

faisalive: sure, added default value for host

Issue #217, opened 2025-09-07
Replace extraction with distillation but reuse the browser profile


yuxicreate:

ariya:
From the current specs, here is the list of classic parsing schema, suitable to target for the migration:

```
$ git grep 'parse:'
amazon/specs.yml:parse:
amazonca/specs.yml:parse:
astro/specs.yml:parse:
audible/specs.yml:parse:
doordash/specs.yml:parse:
ebird/specs.yml:parse:
goodreads/specs.yml:parse:
hardcover/specs.yml:parse:
ikea/specs.yml:parse:
kindle/specs.yml:parse:
lenspure/specs.yml:parse:
linkedin/specs.yml:parse:
nike/specs.yml:parse:
officedepot/specs.yml:parse:
sephora/specs.yml:parse:
thriftbooks/specs.yml:parse:
tokopedia/specs.yml:parse:
ubereats/specs.yml:parse:
zillow/specs.yml:parse:
```

ariya:
Whenever possible, if the extraction/parsing is moved to use distillation, the remove the `extract` and `parse` in the corresponding spec file, to mark the migration.

/cc @faisalive @broerjuang

faisalive:
Other brand can check here: https://github.com/mcp-getgather/middleman/issues/23

PR #232, opened 2025-09-09
feat: replace extraction with distillation but reuse the browser profile

broerjuang:
#217

Reusing the browser profile and using distillation to extract and parse data.

https://github.com/user-attachments/assets/1bbb13e7-eff4-4337-90ea-7bd60c1ddd95

For a separate PR, I think we should:
- Make pattern optional, so we only need to supply the target URL path for extraction

```py
run_distillation_loop(target_url)
```


getgather/distill.py
```diff
@@ -297,14 +298,24 @@ async def distill(hostname: str | None, page: Page, patterns: list[Pattern]) ->
         return match
 
 
-async def run_distillation_loop(location: str, patterns: list[Pattern], fields: list[str] = []):
+async def run_distillation_loop(
+    location: str,
+    patterns: list[Pattern],
+    fields: list[str] = [],
+    browser_profile: BrowserProfile | None = None,
```
broerjuang, line 305: adding optional browser profile

faisalive (approved):

getgather/distill.py
```diff
@@ -297,14 +298,24 @@ async def distill(hostname: str | None, page: Page, patterns: list[Pattern]) ->
...
         raise ValueError("No distillation patterns provided")
 
     hostname = urllib.parse.urlparse(location).hostname or ""
 
-    profile = BrowserProfile()
+    # Use provided profile, or try to get from MCP context, or create new one
+    if browser_profile:
+        profile = browser_profile
+    else:
+        profile = get_mcp_browser_profile() or BrowserProfile()
```
ariya, line 317: `get_mcp_browser_profile` should be invoked by the caller (e.g. `get_book_list()`) since this distillation loop is at the layer where it does not really know about MCP.

getgather/mcp/brand/goodreads.py
```diff
@@ -15,7 +16,12 @@
 @goodreads_mcp.tool(tags={"private"})
 async def get_book_list() -> dict[str, Any]:
     """Get the book list from a user's Goodreads account."""
-    return await extract()
+
+    path = os.path.join(os.path.dirname(os.path.dirname(__file__)), "patterns", "**/*.html")
+    patterns = load_distillation_patterns(path)
+
+    books = await run_distillation_loop("https://www.goodreads.com/review/list", patterns)
```
ariya, line 23: Obtain the `browser_profile` first. If it's empty or default, throw an exception since for now the distillation-based tool can only reuse a "warm" browser profile.

getgather/mcp/brand/goodreads.py
```diff
@@ -17,10 +18,12 @@
 async def get_book_list() -> dict[str, Any]:
     """Get the book list from a user's Goodreads account."""
 
+    browser_profile = get_mcp_browser_profile()
```
broerjuang, line 21: this will throw if no profile is found

getgather/mcp/brand/goodreads.py
broerjuang: updated it on [006e3f6](https://github.com/mcp-getgather/mcp-getgather/pull/232/commits/006e3f626b8ccbd47d7a3aa1bea2c35b1bde3cc1)

getgather/distill.py
```diff
@@ -297,14 +297,21 @@ async def distill(hostname: str | None, page: Page, patterns: list[Pattern]) ->
...
+    browser_profile: BrowserProfile | None = None,
+):
     if len(patterns) == 0:
         logger.error("No distillation patterns provided")
         raise ValueError("No distillation patterns provided")
 
     hostname = urllib.parse.urlparse(location).hostname or ""
 
-    profile = BrowserProfile()
+    # Use provided profile or create new one
```
ariya, line 312: Nitpick: this kind of comment is redundant.

PR #234, opened 2025-09-09
chore: cleanup

broerjuang:
# Standardize Dev Workflow and Add Pre-Push Hooks

This PR cleans up our development scripts and adds some nice dev workflow.

### Package.json Scripts
Reorganized all the npm scripts to use a consistent frontend/backend naming pattern:
- frontend:lint, frontend:check:format, frontend:format
- backend:check:format, backend:format, backend:check:types  
- yaml:check:format
- check:all (runs everything)

Also made some performance improvements:
- Replaced nodemon with chokidar-cli for faster file watching
- Added a conditional predev script that only builds when needed. On the first run, the frontend requires the __assets folder to exist, otherwise the server will throw an error.

### Pre-Push Hooks with Husky
Added automatic code quality checks before pushing:
- Runs the same checks as our CI pipeline locally
- Catches formatting/linting issues before they hit the remote
- Shows helpful error messages with fix commands
- Blocks push if issues are found

### GitHub Actions Updates
Updated the static-analysis.yml workflow to match the new script names:
- Renamed jobs for clarity (python-coding-style -> backend-format, etc)
- Added proper step names instead of bare run commands
- Everything still works the same, just cleaner

## Why These Changes

The old setup had inconsistent naming and no local quality checks. Now developers get:
- Faster development cycles (conditional builds, better file watching)
- Consistent script naming that matches our frontend/backend split
- Quality checks that run locally before push (saves CI time and prevents broken builds)
- Better error messages when things go wrong

## Testing

All the new scripts work and the pre-push hook successfully blocks bad code. The CI pipeline runs the same checks so there's no drift between local and remote validation.

Developers just need to run npm install after this merges to get the git hooks set up automatically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,9 +19,9 @@
     "frontend:lint": "eslint .",
     "backend:check:format": "uv run ruff check && uv run ruff format --check",
     "backend:format": "uv run ruff format && uv run ruff check --fix",
-    "backend:check:types": "uv run pyright .",
+    "backend:check:type-safety": "uv run pyright .",
     "yaml:check:format": "uv run yamlfix --check $(find . -type f -name '*.yml' -o -name '*.yaml' | grep -v -E '\\.venv/|node_modules/')",
-    "check:all": "npm run frontend:lint && npm run frontend:check:format && npm run backend:check:format && npm run backend:check:types && npm run yaml:check:format",
+    "check:all": "npm run frontend:lint && npm run frontend:check:format && npm run backend:check:format && npm run backend:check:type-safety && npm run yaml:check:format",
     "generate:schema": "wait-on tcp:23456 && openapi-typescript http://localhost:23456/api/openapi.json -o frontend/__generated__/api.d.ts",
     "watch:schema:api": "chokidar \"getgather/**/*.py\" -c \"npm run generate:schema\"",
     "prepare": "husky"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,17 +7,24 @@
...
-    "check-format": "prettier --check \"**/*.{html,js,jsx,ts,tsx,css,json,md}\"",
-    "format": "prettier --write \"**/*.{html,js,jsx,ts,tsx,css,json,md}\"",
-    "schema:api": "wait-on tcp:23456 && openapi-typescript http://localhost:23456/api/openapi.json -o frontend/__generated__/api.d.ts",
-    "watch:schema:api": "nodemon --watch getgather -e py --exec \"npm run schema:api\""
+    "frontend:check:format": "prettier --check \"**/*.{html,js,jsx,ts,tsx,css,json,md}\"",
+    "frontend:format": "prettier --write \"**/*.{html,js,jsx,ts,tsx,css,json,md}\"",
+    "frontend:lint": "eslint .",
+    "backend:check:format": "uv run ruff check && uv run ruff format --check",
+    "backend:format": "uv run ruff format && uv run ruff check --fix",
+    "backend:check:types": "uv run pyright .",
```
ariya, line 22: Somehow `check-types` does not convey the same meaning as `typecheck`, in the language compilation sense.
Proposal: call this `backend:check:type-safety` instead.

PR #242, opened 2025-09-10
refactor: doordash orders using distillation 

broerjuang:
eed to check actual doordash account. This changes made following the spec from doordash/spec.yml


ariya:
@broerjuang I tried on my Doordash account, and it looks decent. The parsing isn't great yet, but this is not a regression anyway. Please make sure to remove the `parse` and `extract` block in `getgather/connectors/brand_specs/doordash/specs.yml`.

ariya (approved):

PR #244, opened 2025-09-10
refactor: shopee purchase using distillation

faisalive:
Screenshot 2025-09-10 at 21 33 18


scoutcallens (commented):
How does this work if the user isn't signed into Shoppee? I don't see any htmls for detecting sign in pages.

ariya:
> How does this work if the user isn't signed into Shoppee? I don't see any htmls for detecting sign in pages.

See #217, this is now only working on a "warm" browser profile, post sign in.
The whole thing is replacing extract + parse, not yet a replacement for sign in.

PR #274, opened 2025-09-14
feat - astro with distillation

shabrina-ha:


getgather/mcp/patterns/astro-otp.html
```diff
@@ -0,0 +1,8 @@
+<html gg-domain="astronauts.id">
+  <head>
+    <title>Astro OTP</title>
+  </head>
+  <body>
+    <input type="text" placeholder="One-time code" gg-match='[data-testid="otpInput0"]' />
```
ariya, line 6: This likely needs to optionally match an error message (e.g. when the code is incorrect, or time out).

faisalive (approved):

PR #276, opened 2025-09-15
refactor: move api call to function from mcp

faisalive:
Call function directly from `mcp/shared.py` (create link & status link) instead of fetch API.

Also removing `/home` and `/start/{brand_id}` from Frontend code, since that page consume `/link/create` API.

### DIFF
--- a/getgather/api/routes/link/endpoints.py
+++ b/getgather/api/routes/link/endpoints.py
@@ -1,12 +1,97 @@
+from typing import Annotated
+
 from fastapi import APIRouter, HTTPException, Request, status
 
 from getgather.api.routes.link.types import (
+    HostedLinkTokenRequest,
+    HostedLinkTokenResponse,
     TokenLookupResponse,
 )
+from getgather.connectors.spec_loader import BrandIdEnum
 from getgather.hosted_link_manager import HostedLinkManager, LinkDataUpdate
 from getgather.logs import logger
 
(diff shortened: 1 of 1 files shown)


faisalive:
The tests are failing because this page was deleted.

image

Should we keep this page or remove it entirely? @ariya

broerjuang:
I think we can stick with this for now until we remove the legacy AuthMiddleware and switch to distillation, since our hosted link implementation still depends on it.

getgather/api/routes/link/endpoints.py
```diff
@@ -1,99 +1,14 @@
...
-        raise
-    except Exception as e:
-        logger.error("Error creating hosted link", extra={"error": str(e)}, exc_info=True)
-        raise HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
-            detail="Internal server error while creating hosted link",
-        )
-
-
 @router.get("/status/{link_id}", response_model=TokenLookupResponse)
```
bin-ario, line 12: can this be removed too?

ariya:
yeah, we should not delete `/home` etc until everything is fully migrated.

getgather/api/routes/link/endpoints.py
faisalive: Currently it can't. Still used by hosted link frontend.

PR #283, opened 2025-09-16
fix: auto generate types before run frontend

faisalive:
After #253, `__generated__/api.d.ts` is no longer generated automatically when we run `npm run dev`, which causes type errors.
Now we need to run the `getgather.generate_openapi` script to generate `openapi.json`, and then convert that into `api.d.ts`.

This is causing the frontend build to fail.
Screenshot 2025-09-16 at 14 06 03

After fix:
Screenshot 2025-09-16 at 14 07 53

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,7 +73,7 @@ RUN mkdir -p frontend/__generated__ && \
 COPY frontend/ ./frontend/
 
 # Build frontend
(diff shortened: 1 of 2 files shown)


broerjuang (approved):
LGTM :shipit:

faisalburhanudin (approved):

PR #288, opened 2025-09-16
refactor: extract hardcover using distillation

faisalive:
Screenshot 2025-09-16 at 16 12 28


getgather/mcp/patterns/hardcover-books.html
```diff
@@ -0,0 +1,30 @@
+<html gg-domain="hardcover" gg-priority="1">
+  <body>
+    <div>
+      <div gg-match="a.text-gray-600:has-text('Books')"></div>
```
ariya, line 4: I assume `text-gray-600` has be to specific because otherwise there are too many `Books` in the page?

faisalive: Yes there are many `Books`. We want to detect "active" book tab.

Added more specific selector.

image

ariya (approved):

PR #291, opened 2025-09-16
feat: lenspure distillation 

broerjuang:
Cannot extract lenspure since we cannot reuse browser profile 

Screenshot 2025-09-16 at 16 23 07


faisalive:
Replaced with https://github.com/mcp-getgather/mcp-getgather/pull/455

PR #297, opened 2025-09-17
feat: add quince assets

broerjuang:


getgather/distill.py
```diff
@@ -366,6 +366,8 @@ async def run_distillation_loop(
                     print(distilled)
                     if interactive:
                         await autoclick(page, distilled)
+                        # Give the page time to respond to clicks before next iteration
+                        await asyncio.sleep(2.0)
```
broerjuang, line 370: We need to delay the match a bit since there’s a page navigation after the auto-click. If we run it too quickly, it’ll catch the previous page instead.

faisalive (approved):

PR #303, opened 2025-09-17
feat: add aa distillation

broerjuang:
The jssue regarding reusing browser profile has been fixed since we never really close the browser.

### DIFF
--- a/getgather/mcp/americanairlines.py
+++ b/getgather/mcp/americanairlines.py
@@ -12,5 +12,5 @@ async def get_upcoming_flights() -> dict[str, Any]:
     return await dpage_mcp_tool(
         "https://www.aa.com/aadvantage-program/profile/account-summary",
         "americanairlines_upcoming_flights",
-        timeout=10
+        timeout=10,
     )


faisalive:
Adjusting the pattern, success signin and get the html
Screenshot 2025-10-27 at 14 49 42

getgather/mcp/americanairlines.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@americanairlines_mcp.tool
+async def get_upcoming_flights() -> dict[str, Any]:
+    """Get upcoming flights of americanairlines."""
+    return await dpage_mcp_tool(
+        "https://www.aa.com/aadvantage-program/profile/account-summary",
+        "americanairlines_upcoming_flights",
+        timeout=10,
+    )
```
faisalive: Need to add a `timeout`. If we don’t add this, the `dpage_mcp_tool` will always return a `signin_url` even though we’re already signed in. This happens because the AA site is a SPA (and it’s slow on my connection). Because of that, the distill loop starts immediately (document has loaded, but the content is still loading) and the iteration/timeout is only `2`. So when we only wait for 2 "seconds", the page is still loading and hasn’t yet detected the `gg-stop` pattern.

https://github.com/user-attachments/assets/83d71db0-9e60-4fd1-b297-a3860475211a

Sure, the issue is resolved when we add `timeout=xx`, but the side effect is that the first time we hit it (when not signed in), the response will be slow because it has to wait for the timeout/iteration.

PR #306, opened 2025-09-17
Update patterns folder structure

scoutcallens:
Make the "patterns" folder easier to use, compartmentalizing each brand html into its own folder. All changes are also reflected in the corresponding MCP tools sot he right html files are always found based on the correct brand.


ariya:
Since the benefit is marginal, we should hold off reorganization PR like this, while there are still other in-flight and pending important PRs which migrate the extraction/parse to distillation (otherwise, those PRs need to be reworked by the author to adopt the directory structure).

PR #313, opened 2025-09-18
refactor: ebay to distillation

faisalive:
Screenshot 2025-09-18 at 14 02 27


getgather/mcp/patterns/ebay-order-history.html
```diff
@@ -0,0 +1,9 @@
+<html gg-domain="goodreads">
```
scoutcallens, line 1: Update the titles here to eBay

PR #335, opened 2025-09-19
refactor: nordstorm distillation (extract, mcp_tool)

broerjuang:
Screenshot 2025-09-19 at 16 31 11


getgather/mcp/patterns/nordstrom-orders.html
```diff
@@ -0,0 +1,10 @@
+<html gg-domain="nordstrom">
+  <head>
+    <title>Nordstrom Orders</title>
+  </head>
+  <body>
+    <div>
+      <div gg-stop gg-match-html="div.NsJwo"></div>
```
broerjuang, line 7: seems like all className is generated. So it can't be more generic than this.

PR #336, opened 2025-09-19
refactor: migrate officedepot to use distillation

oshimayoan:
Please do not merge this PR yet before the migration in data-portrait and return-reminder for this changes are ready.

https://github.com/user-attachments/assets/e66280f0-c811-480e-a428-7cd6893b6ae1

### DIFF
--- a/getgather/mcp/brand/officedepot.py
+++ b/getgather/mcp/brand/officedepot.py
@@ -38,4 +38,8 @@ async def get_order_history_details(order_number: str) -> dict[str, Any]:
         browser_profile=browser_profile,
     )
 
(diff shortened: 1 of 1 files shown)


oshimayoan:
@ariya Yes, but I've discussed it with @broerjuang before and he said that he will close his branch. This diff is covering more things as the order depot UI has been redesigned, I think. There is no product name anymore in their order page unless we click on an order to navigate to the order detail page, similar with wayfair UI.

ariya:
> @ariya Yes, but I've discussed it with @broerjuang before and he said that he will close his branch.

In that case, @broerjuang please close PR #243.

PR #386, opened 2025-09-22
Remove inspector endpoint

### LINKED ISSUE #349: Remove the built-in MCP inspector
It's an additional maintenance burden with marginal benefit, and hard to make it work with mcp-gateway (including its auth).

Any developer can always launch the inspector externally with `npx @modelcontextprotocol/inspector` and then connect to the right MCP server.

scoutcallens:
As titled.

Closes #349


ariya:
@scoutcallens See also PR #148.

PR #397, opened 2025-09-23
Add proxy to MCP tests

scoutcallens:
Adds the proxy endpoint and password to the MCP tests so that certain MCP tests don't flake out due to the CI using its own (probably detectable) IP address. Should make the NYT bestsellers test less flaky if it works properly. However, if this runs on every PR then we might want to think about cost. I'm not sure how expensive it is to run the proxy but it might not be worth it to run for every MCP test.


scoutcallens:
Closing for now as it made it worse instead of better.

Issue #395, opened 2025-09-23
MCP tools file order issue


faisalive:
Facing an issue when placing an MCP file (`{brand}.py`) in `/getgather/mcp/`.

It turns out that the file order matters. The issues are as follows:

- If the file below `registry.py`. The MCP tools are not listed, even though the MCP registration/mount is successful.
As shown in the screenshot below, the Sephora MCP is successfully registered, but its tools do not appear in the MCP tools list.

<img  height="250" alt="Image" src="https://github.com/user-attachments/assets/d38a889c-7028-4a10-88c5-1cd2ca8a97a5" />

##

- If the file below `dpage.py`, It always shows `{"detail":"Invalid page id"}` when opening `http://localhost/dpage/{id}`.
It turns out the `active_pages` value is always empty when the URL is opened.

<img width="506" height="283" alt="Image" src="https://github.com/user-attachments/assets/1f5ff903-4f06-4d47-8fd2-e072b2ce5c27" />

For testing purpose you can use this PR branch (#394). And try to rename the `sephora.py` file to `psephora.py` (above `registry.py`) or `csephora.py` (above `dpage.py`)

PR #398, opened 2025-09-23
Fix autoimport by checking for the MCP class in the module source

ariya:
Eventually when everything is migrated to distillation, we may not need this. But for now, avoid blind module import by checking first heuristically if the module is a legimate MCP class.

To verify, run the MCP server, connect MCP inspector, list some tools. All the available tools should be still enumerated correctly.


faisalive (approved):
close #395

PR #400, opened 2025-09-24
fix - use http when running local

shabrina-ha:


faisalive:
@shabrina-ha are these changes identical to #399?

PR #413, opened 2025-09-26
add the ability to use oxylabs proxy directly

bin-ario:
we can use `http://:1234@localhost:8889` for our own proxy service, where username is dynamic
or `https://username:password@pr.oxylabs.io:7777` for oxylabs directly
or any other proxy services

### DIFF
--- a/getgather/browser/proxy.py
+++ b/getgather/browser/proxy.py
@@ -4,6 +4,11 @@ This module provides proxy configuration for external proxy service integration
 with hierarchical location support (city, state, country).
 """
 
+import re
+from urllib.parse import urlsplit
+
+from patchright.async_api import ProxySettings
+
 from getgather.api.types import RequestInfo
 from getgather.config import settings
 from getgather.logs import logger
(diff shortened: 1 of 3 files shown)


ariya:
First make sure that the env for the deployed MCP server already adopts the new `BROWSER_PROXY`.

getgather/browser/proxy.py
```diff
@@ -28,12 +33,23 @@ async def setup_proxy(
...
-    if not settings.BROWSER_HTTP_PROXY or not settings.BROWSER_HTTP_PROXY_PASSWORD:
-        logger.info(
-            "No proxy configured (BROWSER_HTTP_PROXY and BROWSER_HTTP_PROXY_PASSWORD not set)"
-        )
+    if not settings.BROWSER_PROXY:
         return None
 
+    proxy = urlsplit(settings.BROWSER_PROXY)
+    if not proxy.password:
+        raise ValueError("BROWSER_PROXY must contain a password")
```
ariya, line 41: Can't password be optional? I imagine some internal proxy server (not for public) which does not require any password.

ariya (approved):
LGTM, a minor clarification.

bin-ario:
doesn't seem useful anymore

PR #424, opened 2025-09-30
docs: add toml example for codex

adamjsawicki:
## Context

As I was going through, realized the docs were missing a section for codex, so I added.


ariya (approved):
Thank you @adamjsawicki

PR #493, opened 2025-10-09
fix: changing the parse_html param to string instead of brandenum 

broerjuang:
The idea is to handle the migration one step at a time. Since our tool still depends on the previous one (like parse_html), we’ll need to keep supporting it until we’re fully migrated to DPage.

Working with `search_product`
Screenshot 2025-10-09 at 16 43 14

Previously
Screenshot 2025-10-09 at 16 44 05

still need to test private tool


broerjuang:
I’m okay with using an enum. However, this PR is related to [#475](https://github.com/mcp-getgather/mcp-getgather/pull/475), where the idea is to refactor the previous spec into Distillation.

The issue is that there’s a conflict between how GatherMCP and BrandMCPBase resolve BrandEnum.

cc: @kpprasa

PR #509, opened 2025-10-11
Use a global session

kpprasa:
This is the start of an extensive refactor to more properly use sessions.


ariya:
Closing this for now, as likely it'll be easier to handle with the upcoming Zendriver migration (see #608).

PR #517, opened 2025-10-15
Properly create a new page from the browser context

ariya:
This is to account for resource filtering, see PR #503 and #516.

Without calling `context.new_page()`, the "old" page without resource filtering will be used, which has all sorts of loading issues (too much extra trackers and ads etc).

This applies to both the sign-in orchestrator and distillation.

To verify, launch as usual to /mcp-shopping and /mcp-media. For the former, try the Amazon tool (sign-in orchestrator) and for the latter, use the BBC tool (distillation).

Before this PR, opening the hosted link/dpage will load Amazon/BBC with its images and other heavy resources, indicating that resources filtering did not apply to that page.

After this PR, that initial page of Amazon/BBC has no images at all (as well as custom fonts, etc, including trackers and ads). This is the ultimate proof that resources are filtered properly.

2025-10-14 amazon sans images

### DIFF
--- a/getgather/browser/session.py
+++ b/getgather/browser/session.py
@@ -53,10 +53,9 @@ class BrowserSession:
 
     async def page(self) -> Page:
         # TODO: It's okay for now to return the last page. We may want to track all pages in the future.
-        if self._context is not None:
-            if self._context.pages and len(self._context.pages) > 0:
-                logger.info(f"Returning existing page in context with profile {self.profile.id}")
-                return self._context.pages[-1]
+        if self.context.pages and len(self.context.pages) > 0:
+            logger.info(f"Returning existing page in context with profile {self.profile.id}")
+            return self.context.pages[-1]
         return await self.new_page()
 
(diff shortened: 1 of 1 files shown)


getgather/browser/session.py
```diff
@@ -47,11 +47,17 @@ def playwright(self) -> Playwright:
...
+    async def new_page(self) -> Page:
+        logger.info(f"Creating new page in context with profile {self.profile.id}")
+        return await self.context.new_page()
+
     async def page(self) -> Page:
         # TODO: It's okay for now to return the last page. We may want to track all pages in the future.
-        if not self.context.pages:
-            await self.context.new_page()
-        return self.context.pages[-1]
+        if self._context is not None:
```
bin-ario: the intention to call `self.context` instead of `self._context` was to raise exception if context/session is not initialized correctly, seems like it still makes sense here

bin-ario (approved):
LGTM

PR #522, opened 2025-10-15
handle sticky dpage browser profile for demo app

faisalive:
We want to use the `x-incognito` header in our demo app so it always launche a fresh browser profile for every MCP call ([#34](https://github.com/mcp-getgather/data-portrait/issues/34)).

However, some brands require multiple MCP tool calls (e.g., [Wayfair](https://github.com/mcp-getgather/data-portrait/blob/main/src/server/handlers/mcp-handler.ts#L14), which needs to fetch both list and details). Currently, there are no issue because we still use the `AuthOrchestrator` for Wayfair. But once we migrate the signin flow to dpage and use `x-incognito`, it will cause a problem, each MCP tool call will always trigger a new signin since it always use fresh browser profile.

So I propose creating a “sticky” browser profile for our demo app, mapped to the MCP session ID (similar to how we previously handled it in [BrandState](https://github.com/mcp-getgather/mcp-getgather/blob/main/getgather/mcp/brand_state.py#L28-L47))

### DIFF
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -7,7 +7,7 @@ from typing import Any
 from bs4 import BeautifulSoup, Tag
 from fastapi import APIRouter, HTTPException, Request
 from fastapi.responses import HTMLResponse
-from fastmcp.server.dependencies import get_http_headers
+from fastmcp.server.dependencies import get_context, get_http_headers
 from nanoid import generate
 from patchright.async_api import Page
 
@@ -33,6 +33,7 @@ router = APIRouter(prefix="/dpage", tags=["dpage"])
 active_pages: dict[str, Page] = {}
 distillation_results: dict[str, str | list[dict[str, str | list[str]]]] = {}
 global_browser_profile: BrowserProfile | None = None
+browser_profiles: dict[str, BrowserProfile] = {}
 
 
 async def dpage_add(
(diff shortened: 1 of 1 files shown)


faisalive:
@ariya @broerjuang, we’re having an issue with this PR. Since we now only run [run_distillation_loop](https://github.com/mcp-getgather/mcp-getgather/blob/main/getgather/mcp/dpage.py#L344-L354)
 when `incognito=False`, we still need a flag to run that function even when `incognito=True`.

In #557 and #558, we can have extra conditions like [this](https://github.com/mcp-getgather/mcp-getgather/pull/557/files#diff-c5724c0feeaae70325079415f39735174f4aafe2443e9edc4378b8994dc43aceR349) and [this](https://github.com/mcp-getgather/mcp-getgather/pull/558/files#r2449839207) because we have flag/variable that indicates the client has already signed in and wants to run `run_distillation_loop`.

If we don’t add this flag, every time we call the tool, it will always return the dpage URL.

So I think using the approach from #558 makes more sense, although we’ll need to adjust the code in the demo apps.

broerjuang:
Just to make sure I’m following:

> So I think using the approach from [PR #558](https://github.com/mcp-getgather/mcp-getgather/pull/558) makes more sense, although we’ll need to adjust the code in the demo apps.

Here’s how it works: when a user initiates an MCP tool call, it will return a signin-id. On the second call, this value will be included in the header as x-signin-id. @faisalive

PR #526, opened 2025-10-16
feat: dpage with manual control for custom actions

broerjuang:
Rationale:
We currently have two ways of doing things — the old FSM orchestration and DPage. The thing is, DPage relies a lot on patterns, which makes it tricky to handle manual steps like “go to this page first, do something, extract this, then parse that.”

This experiment adds a callback that passes the page and session, so we can manually define what happens after each action and handle more flexible flows.

Screenshot 2025-10-16 at 17 25 36

The problem of current approach:

Screenshot 2025-10-16 at 17 43 46

The Progress:
Screenshot 2025-10-17 at 14 37 23

### DIFF
--- a/getgather/mcp/brand/goodreads.py
+++ b/getgather/mcp/brand/goodreads.py
@@ -1,32 +1,14 @@
-import os
 from typing import Any
 
 from fastmcp import Context
 
-from getgather.distill import load_distillation_patterns, run_distillation_loop
 from getgather.mcp.agent import run_agent_for_brand
 from getgather.mcp.registry import BrandMCPBase
-from getgather.mcp.shared import get_mcp_browser_profile
 from getgather.mcp.stagehand_agent import (
     run_stagehand_agent,
 )
 
 goodreads_mcp = BrandMCPBase(brand_id="goodreads", name="Goodreads MCP")
-
-
-@goodreads_mcp.tool(tags={"private"})
-async def get_book_list() -> dict[str, Any]:
-    """Get the book list from a user's Goodreads account."""
-
-    browser_profile = get_mcp_browser_profile()
-    path = os.path.join(os.path.dirname(os.path.dirname(__file__)), "patterns", "**/*.html")
-    patterns = load_distillation_patterns(path)
-    books, _ = await run_distillation_loop(
-        "https://www.goodreads.com/review/list?ref=nav_mybooks&view=table",
-        patterns,
-        browser_profile=browser_profile,
-    )
-    return {"books": books}
 
 
 @goodreads_mcp.tool(tags={"private"})
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -36,22 +36,15 @@ router = APIRouter(prefix="/dpage", tags=["dpage"])
 active_pages: dict[str, Page] = {}
 distillation_results: dict[str, str | list[dict[str, str | list[str]]] | dict[str, Any]] = {}
 pending_actions: dict[str, dict[str, Any]] = {}  # Store actions to resume after signin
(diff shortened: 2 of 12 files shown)


broerjuang:
> dpage_mcp_tool function is only 50 lines. It seems to be quite overkill to branch its internal behavior based on a potential callback. Afterall, it's intended as a shortcut so that a class of MCP tools do not need to repeat the logic.

> Here's an alternative: create a new variant called dpage_action_mcp_tool. The logic is similar: run a parametrized "manual" action, and based on that result (action is successful or not), perform dpage-style sign-in.

Noted, will do this one. I’ve been thinking about using a different function, but since a lot of it was redundant, I decided to reuse the same function for this experiment. I’ll refactor it once it’s proven to work.

getgather/mcp/brand/goodreads.py
```diff
@@ -1,32 +1,39 @@
...
+        await page.click("a:has-text('My Books')")
+        return {"url": page.url}
+
+    return await dpage_mcp_tool(
+        "https://www.goodreads.com/recommendations",
+        "url",
+        callback=callback,
+    )
+
+
```
broerjuang: this is the api if you pass callback. It accepts page as first argument and you can do manual work here. Just assume that you're on the `initial_page` already.

getgather/mcp/brand/goodreads.py
```diff
@@ -1,32 +1,39 @@
...
-@goodreads_mcp.tool(tags={"private"})
+@goodreads_mcp.tool
+async def get_username() -> dict[str, Any]:
+    async def callback(page: Page) -> dict[str, Any]:
+        await page.click("a:has-text('My Books')")
+        return {"url": page.url}
+
+    return await dpage_mcp_tool(
+        "https://www.goodreads.com/recommendations",
+        "url",
```
broerjuang: the second argument is irrelevant here. Since the value being returned is coming from line 20 above. Will refactor this when we split dpage_mcp_tool into its own function

getgather/mcp/dpage.py
```diff
@@ -49,7 +50,7 @@ async def dpage_add(
     if browser_profile is None:
         browser_profile = BrowserProfile()
 
-    session = BrowserSession(browser_profile.id)
+    session = BrowserSession.get(browser_profile)
```
faisalive: why this changed?

getgather/mcp/dpage.py
```diff
@@ -325,9 +400,19 @@ async def dpage_mcp_tool(initial_url: str, result_key: str, timeout: int = 2) ->
...
     id = await dpage_add(browser_profile=browser_profile, location=initial_url)
 
+    # Store callback for auto-resumption after signin
+    if callback is not None:
+        pending_callbacks[id] = {
+            "callback": callback,
+            "initial_url": initial_url,
+            "result_key": result_key,
+            "timeout": timeout,
+            "page_id": id,
```
faisalive: why we need to store `page_id` here?

getgather/mcp/dpage.py
```diff
@@ -339,9 +383,14 @@ async def dpage_mcp_tool(initial_url: str, result_key: str, timeout: int = 2) ->
 
     url = f"{base_url}/dpage/{id}"
     logger.info(f"Continue with the sign in at {url}", extra={"url": url, "id": id})
+
+    message = "Continue to sign in in your browser"
+    if callback is not None:
+        message = "Manual action failed. Continue to sign in"
+
```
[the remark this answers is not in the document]
broerjuang: it's introduced

ariya (approved):
LGTM, ship it!

PR #530, opened 2025-10-17
Use a tuple to indicate termination in the distillation loop

ariya:
Rather than using the terminate flag (see PR #422), this adds the use of idiomatic tuple to always supply the terminated flag (up to the caller to use it or not).

I apologize for breaking pending/upcoming PRs for distillation migration.


broerjuang (approved):
Nice, it's more explicit! Been thinking about using NewType but it will be too much for now. Ship it

PR #532, opened 2025-10-17
close dpage browser after terminate

faisalive:
Because we never close the browser after the dpage reaches the `terminate` state, there are a lot of "zombie" browsers running on our machine.

image

### DIFF
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -65,7 +65,7 @@ async def dpage_add(
 
 async def dpage_close(id: str) -> None:
     if id in active_pages:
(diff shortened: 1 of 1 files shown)


ariya:
Oh, actually now remember the reason: it doesn't really matter much if the page is closed or now, because it will be reused again for any next call to the page.

The drawback of not closing ti: the page is not really idle, and its activities might actually consume the bandwidth.

faisalive:
> Oh, actually now remember the reason: it doesn't really matter much if the page is closed or now, because it will be reused again for any next call to the page.

If we use `x-incognito`, I think it won’t be reused, right? @ariya

PR #535, opened 2025-10-17
Remove page reuse

kpprasa:
I think I found a reason for the failing tests. 

We used page, but we should use new_page whenever using the global session.


kpprasa:
@ariya, I agree. I was initially worried about changing old code, but I took a look and I don't think any of those uses depended on getting the last page.

kpprasa:
- [ ] https://github.com/mcp-getgather/mcp-getgather/pull/536

PR #536, opened 2025-10-17
Make all page calls new page

kpprasa:
In the era of a sometimes global session, it is far safer.


kpprasa:
Closing as not worth the risk

PR #555, opened 2025-10-20
Reuseable short-lived MCP tool

ariya:
This is to abstract the tool used for NPR (#541), CNN (#543), and Ground News (#545).

### DIFF
--- a/getgather/distill.py
+++ b/getgather/distill.py
@@ -8,6 +8,7 @@ from datetime import datetime
 from glob import glob
 from pathlib import Path
 from typing import Any, cast
+from urllib.parse import urlparse, urlunparse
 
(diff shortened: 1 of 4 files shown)


getgather/distill.py
```diff
@@ -575,3 +576,41 @@ async def run_distillation_loop(
...
+                    url: str = urlunparse((
+                        "https",
+                        netloc,
+                        parsed.path,
+                        parsed.params,
+                        parsed.query,
+                        parsed.fragment,
+                    ))
+                    item["url"] = url
+    return terminated, result
```
bin-ario: why not raise exception here if not terminated?

bin-ario (approved):

PR #558, opened 2025-10-21
reusable incognito browser (alternative)

faisalive:
Alternative way of #557 . Instead of returning the `browser_profile_id`, we can store that by signin_id

### DIFF
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -34,7 +34,7 @@ router = APIRouter(prefix="/dpage", tags=["dpage"])
 
 
 active_pages: dict[str, Page] = {}
-browser_profiles: dict[str, BrowserProfile] = {}
+incognito_browser_profiles: dict[str, BrowserProfile] = {}
 distillation_results: dict[str, str | list[dict[str, str | list[str]]]] = {}
 global_browser_profile: BrowserProfile | None = None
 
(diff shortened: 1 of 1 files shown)


faisalive:
> If it's not easy/possible to send a custom header for calling MCP tool, then I think this PR will not work anyway, so just close it for now.

Yes, currently it's not possible to send custom header via tool call. But it's still possible to re-connect the MCP client with different custom headers (`x-signin-id`), then this PR should work.

> As for #522, how's different than #557? If it's similar, is it possible to close both and come up with a new PR that merge the concept?

For #522 we're using `session_id` that already provided by fastmcp, this is unique ID that we get when we conect mcp client to a server, so we don't need to re-connect the MCP client.

#557 are same as this PR (#558), we need to re-connect the MCP client.

--- 
Or, for other alternative, we can use Grabbit unique id as the identifier (we send that via custom headers) https://github.com/gather-engineering/grabbit/issues/34

ariya:
> For #522 we're using `session_id` that already provided by fastmcp, this is unique ID that we get when we conect mcp client to a server, so we don't need to re-connect the MCP client.

If MCP offers `session_id` and this avoids reconnecting the MCP client, then I think we should try this first.

faisalive:
> > For #522 we're using `session_id` that already provided by fastmcp, this is unique ID that we get when we conect mcp client to a server, so we don't need to re-connect the MCP client.
> 
> If MCP offers `session_id` and this avoids reconnecting the MCP client, then I think we should try this first.

@ariya unfortunately that PR still have problem, see https://github.com/mcp-getgather/mcp-getgather/pull/522#issuecomment-3430671227

PR #562, opened 2025-10-23
exclude amazon.ca from blocklist

faisalive:
It turns out that amazon.ca is in our blocklist

https://raw.githubusercontent.com/hectorm/hmirror/master/data/mozilla-shavar-advertising/list.txt
image


ariya (approved):
LGTM, with suggested modifications.

Issue #552, opened 2025-10-20
Distillation result: supply both raw HTML and JSON


ariya:
Right now the distillation returns either HTML or JSON. For debugging purposes, let's supply both now.

PR #564, opened 2025-10-23
feat: extend distillation return type

broerjuang:
# Extend distillation return tuple structure

## Changes

Extended `run_distillation_loop` return value from 2 elements to 3 elements:

**Before:**
```python
result, terminated = await run_distillation_loop(...)
# result: converted JSON or raw HTML (ambiguous)
# terminated: bool
```

**After:**
```python
converted_or_raw, raw_html, terminated = await run_distillation_loop(...)
# converted_or_raw: converted JSON if successful, otherwise raw HTML
# raw_html: always the raw distilled HTML
# terminated: bool indicating successful termination
```

#552


getgather/mcp/brand/aliexpress.py
```diff
@@ -14,7 +14,8 @@ async def get_orders() -> dict[str, Any]:
     browser_profile = get_mcp_browser_profile()
     path = os.path.join(os.path.dirname(os.path.dirname(__file__)), "patterns", "**/*.html")
     patterns = load_distillation_patterns(path)
-    orders, _ = await run_distillation_loop(
+    _terminated, distilled, converted = await run_distillation_loop(
         "https://www.aliexpress.com/p/order/index.html", patterns, browser_profile=browser_profile
     )
+    orders = converted if converted else distilled
```
broerjuang: It' up to the client what will be sent.

getgather/mcp/dpage.py
```diff
@@ -360,6 +360,7 @@ async def dpage_mcp_tool(initial_url: str, result_key: str, timeout: int = 2) ->
             stop_ok=False,  # Keep global session alive
         )
         if terminated:
+            distillation_result = converted if converted else distilled
```
broerjuang: mantain backeward compatibility for dpage

getgather/distill.py
```diff
@@ -487,7 +487,14 @@ async def run_distillation_loop(
     timeout: int = 15,
     interactive: bool = True,
     stop_ok: bool = False,
-) -> tuple[dict[str, str | ConversionResult | None] | str | ConversionResult, bool]:
+) -> tuple[bool, str, ConversionResult | None]:
```
broerjuang: change the return type to `terminated, distilled, converted`

getgather/mcp/brand/amazon.py
```diff
@@ -41,12 +41,13 @@ async def get_purchase_history(
...
     patterns = load_distillation_patterns(path)
-    purchases, _ = await run_distillation_loop(
+    _terminated, distilled, converted = await run_distillation_loop(
         f"https://www.amazon.com/your-orders/orders?timeFilter=year-{target_year}&startIndex={start_index}",
         patterns,
         browser_profile=browser_profile,
         stop_ok=True,
     )
+    purchases = converted if converted else distilled
     return {"purchases": purchases}
```
ariya: While this is fine for now, I think eventually we want to be explicit to the caller.
Perhaps a future follow-up will need two keys for the return object, `purchases` on a successfully converted data, and something else (e.g. `raw_data` or a better name) if the conversation to JSON fails and thus this will contain the original HTML fragment.

PR #572, opened 2025-10-28
feat: separated github action for all-getgather deployment

faisalburhanudin:
The all-getgather instance, used for testing, is currently blocked from testing changes because the deploy-dokku.yml workflow has been disabled. To resolve this, we need to separate the GitHub Actions workflow so that all-getgather can be deployed and released independently.


ariya (approved):
Looks good, with one remark.

PR #591, opened 2025-10-30
add simple retry to page.goto

bin-ario:
override page.goto with simple retry loop

this should help with sentry errors like
```
TimeoutError
Page.goto: Timeout 30000ms exceeded.
Call log:
  - navigating to "https://www.goodreads.com/review/list?ref=nav_mybooks&view=table", waiting until "load"
```

simulated test log look like
```
[10/30/25 10:59:33] WARNING  page.goto http://localhost:23456/test 1 of 3 failed: Page.goto: Timeout 3000ms
                             exceeded.                                                                     
                             Call log:                                                                     
                               - navigating to "http://localhost:23456/test", waiting until "load"         
                                                                                                           
                             Retrying...  
```

### DIFF
--- a/getgather/browser/session.py
+++ b/getgather/browser/session.py
@@ -4,7 +4,7 @@ import asyncio
 from collections import defaultdict
 from contextlib import asynccontextmanager, suppress
 from datetime import datetime
-from typing import ClassVar
+from typing import ClassVar, Literal
 
 from fastapi import HTTPException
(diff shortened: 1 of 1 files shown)


bin-ario:
> Willl this switch to a different/rotating proxy?

not in this PR. it only solves the problem when goodreads itself is not performing

PR #601, opened 2025-11-04
feat - retry connection

shabrina-ha:
If there's any error of:
- "ERR_CONNECTION_CLOSED"
- "ERR_TUNNEL_CONNECTION_FAILED"
- "ERR_PROXY_CONNECTION_FAILED"
We'll retry the connection, so that it'll minimize failure.


ariya:
Let's hold this off until #599 is tested on production (merged, but not yet deployed).
Assuming the initial connection is verified correctly (with an internal retry), then subsequent remedy might not be necessary.

ariya:
@shabrina-ha  I think we can close this one for the time being, since

* the initial connection is verified already, as in PR #599 
* the attempt to reconnect with `goto` will be taken care of in #591

PR #604, opened 2025-11-05
Use tested incognito browser session for short-lived MCP tools

ariya:
This is a follow-up to PR #599, see that PR for the details.

Note that those short-lived MCP tools are for NPR, CNN, etc.


getgather/mcp/dpage.py
ariya: Moved to `distill.py` since it's a generic function, not only for distilled page.

faisalive (approved):

PR #609, opened 2025-11-05
Avoid root and user a regular user in the container

ariya:
This means we can avoid `--no-sandbox` when launching Chromium. There won't be anymore complain from Chromium because of that.

Avoiding `--no-sandbox` also restores and reenables Chromium security features (with multi-process etc).

See previously https://github.com/mcp-getgather/middleman/pull/57.


ariya:
Not sure I can fix this quickly, so in the mean time I separate the container user first in PR #611.

PR #616, opened 2025-11-07
fix - american airlines distillation based

shabrina-ha:
Signed in succesfully:
CleanShot 2025-11-07 at 07 22 26@2x

Sample of empty upcoming trips:
CleanShot 2025-11-07 at 07 25 33@2x


getgather/mcp/americanairlines.py
```diff
@@ -9,8 +12,55 @@
 @americanairlines_mcp.tool
 async def get_upcoming_flights() -> dict[str, Any]:
     """Get upcoming flights of americanairlines."""
-    return await dpage_mcp_tool(
+
+    async def action(page: Page) -> dict[str, Any]:
+        try:
+            async with page.expect_response(
```
broerjuang: I think we can simplify this by using `handle_network_extraction` from `actions.py`


```py
async def action(page: Page) -> dict[str, Any]:
  resp = handle_network_extraction(page, "/api/upcoming-trips")
  ....
```

getgather/mcp/americanairlines.py
```diff
@@ -9,8 +12,55 @@
...
+                lambda resp: "api/loyalty/miles/transaction/orchestrator/memberActivity"
+                in resp.url,
+                timeout=15000,
+            ) as response_info:
+                await page.goto(
+                    "https://www.aa.com/aadvantage-program/profile/account-summary",
+                    wait_until="commit",
+                )
```
broerjuang: this too

getgather/mcp/main.py
```diff
@@ -115,7 +115,14 @@ async def on_call_tool(self, context: MiddlewareContext[Any], call_next: CallNex
...
     "books": ["goodreads"],
-    "shopping": ["amazon", "amazonca", "shopee", "wayfair", "astro"],
+    "shopping": [
+        "amazon",
+        "amazonca",
+        "shopee",
+        "wayfair",
+        "astro",
+        "americanairlines",
+    ],
```
faisalive: Maybe you can add a new category instead of adding aa to shopping.

broerjuang: I think this one doesn't need new category either. It can be only be viewed inside /mcp

faisalburhanudin (approved):

PR #637, opened 2025-11-11
feat(dpage): officedepot purchase history detail

oshimayoan:


broerjuang:
anyway since the detail has been removed should we also remove `mcp/brand` and spec for officedepot? @oshimayoan

PR #638, opened 2025-11-12
Migrate amazon and amazonca to dpage

shabrina-ha:
**dpage_get_purchase_history**
CleanShot 2025-11-12 at 13 06 38@2x
**get_purchase_history**
CleanShot 2025-11-12 at 13 07 53@2x
**search_purchase_history**
CleanShot 2025-11-12 at 13 10 25@2x
**search_product**
CleanShot 2025-11-12 at 13 12 09@2x
**get_browsing_history**
CleanShot 2025-11-12 at 13 31 56@2x

**Related PR to merge:**
- [return reminder](https://github.com/mcp-getgather/return-reminder/pull/57)

### DIFF
--- a/getgather/mcp/main.py
+++ b/getgather/mcp/main.py
@@ -107,7 +107,7 @@ class AuthMiddleware(Middleware):
 CATEGORY_BUNDLES: dict[str, list[str]] = {
     "food": ["doordash", "ubereats"],
     "books": [],
-    "shopping": ["shopee", "astro"],
+    "shopping": ["amazon", "shopee", "astro"],
     "media": [],
 }
 


getgather/mcp/dpage.py
```diff
@@ -375,7 +375,7 @@ async def dpage_mcp_tool(initial_url: str, result_key: str, timeout: int = 2) ->
             stop_ok=False,  # Keep global session alive
         )
         if terminated:
-            distillation_result = converted if converted else distilled
+            distillation_result = converted if converted is not None else distilled
```
broerjuang: why it needs to be explicitly declared?

shabrina-ha: I tested with account with no orders or no browser history, without explicitly declaring this, empty array is not returned. It'll return html, which works fine with chat, but with return reminder for example, it'll give error

getgather/mcp/persist.py
```diff
@@ -73,49 +73,54 @@ def get(self, key: str) -> T | None: ...
...
+        with self._lock:
+            self.load()
+            return self.rows
 
     def reset(self) -> None:
-        self._indexes = {}
-        self.rows = []
+        with self._lock:
+            self._indexes = {}
+            self.rows = []
```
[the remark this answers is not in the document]
shabrina-ha: This is an attempt to fix the tests that failed [here](https://github.com/mcp-getgather/mcp-getgather/actions/runs/19282111159/job/55135249874). So that there won't be race condition. I can separate this into different PR of course if needed

PR #639, opened 2025-11-12
fix: nordstrom MFA dpage UI

faisalive:
Before:
Screenshot 2025-11-12 at 12 00 16

After:
Screenshot 2025-11-12 at 11 59 25

Nordstrom UI:
Screenshot 2025-11-12 at 11 57 54


broerjuang (approved):
Nice, somehow we should document this UI somewhere

PR #642, opened 2025-11-13
feat(amazon): add new pattern for otp using phone number

broerjuang:
Screenshot 2025-11-13 at 09 44 38

### DIFF
--- a/getgather/mcp/patterns/amazon-otp-phone.html
+++ b/getgather/mcp/patterns/amazon-otp-phone.html
@@ -3,10 +3,10 @@
     <title>Amazon OTP</title>
   </head>
   <body>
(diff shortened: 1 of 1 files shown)


getgather/mcp/patterns/amazon-otp-phone.html
```diff
@@ -0,0 +1,15 @@
+<html gg-domain="amazon">
+  <head>
+    <title>Amazon OTP</title>
+  </head>
+  <body>
+    <span gg-match="div#channelDetailsForOtp span:has-text('your phone')"></span>
+    <h2 gg-match="span:has-text('Enter verification code')">Enter verification code</h2>
```
faisalive: Maybe we can switch these two lines so the UI looks better.

PR #657, opened 2025-11-17
migrate amazon to dpage fully

shabrina-ha:


getgather/mcp/main.py
```diff
@@ -107,7 +107,7 @@ async def on_call_tool(self, context: MiddlewareContext[Any], call_next: CallNex
 CATEGORY_BUNDLES: dict[str, list[str]] = {
     "food": ["doordash", "ubereats"],
     "books": [],
-    "shopping": ["amazon", "astro"],
+    "shopping": ["astro"],
```
broerjuang: Image

I think We also need to change GETHER_URL on our demo app.

PR #659, opened 2025-11-17
extract all amazon browsing history

faisalive:
Extract amazon order history by hitting the endpoint for getting browsing history data. Then parse the response using distillation convert.

Screenshot 2025-11-17 at 18 47 08

### DIFF
--- a/getgather/mcp/amazon.py
+++ b/getgather/mcp/amazon.py
@@ -1,12 +1,13 @@
 import asyncio
+import json
 import os
 from datetime import datetime
-from typing import Any
+from typing import Any, cast
 
-from patchright.async_api import Page
+from patchright.async_api import Page, Response
 
 from getgather.browser.profile import BrowserProfile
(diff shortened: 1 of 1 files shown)


faisalive:
> Does this already do the `PageDown` a few times to scroll to the end? The browsing history is a form of infinite scrolling, so some items are loaded just in time, if there are many of them.

@ariya  This approach doesn’t use `PageDown`/scrolling. Scrolling only loads 8 items per page, so it will take a long time if we have, let’s say, hundreds of items.

So instead of scrolling, I directly call the endpoint to fetch the data with 100 items per request ([here](https://github.com/mcp-getgather/mcp-getgather/pull/659/files#diff-c6cf16cd7cc14f65f44bbab1b401962894e3cfb3c232768ee42a8b1b6ed15cddR152-R154)) (I’ve tested with more than 100 items per request and received an error from Amazon). The example below retrieves about 150 items in only ~9 seconds.

ariya (approved):

PR #663, opened 2025-11-18
fix: separate stop_ok flag on distillation loop

faisalive:
`stop_ok` means to stop the session, not the page. If we only need to stop or close the page (but not the session), we need a new flag

### DIFF
--- a/getgather/mcp/amazon.py
+++ b/getgather/mcp/amazon.py
@@ -63,9 +63,7 @@ async def search_product(keyword: str) -> dict[str, Any]:
 async def get_browsing_history() -> dict[str, Any]:
     """Get browsing history from amazon."""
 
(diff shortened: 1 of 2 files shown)


getgather/mcp/amazon.py
```diff
@@ -63,9 +63,7 @@ async def search_product(keyword: str) -> dict[str, Any]:
 async def get_browsing_history() -> dict[str, Any]:
     """Get browsing history from amazon."""
 
-    async def get_browsing_history_action(
-        page: Page, browser_profile: BrowserProfile
-    ) -> dict[str, Any]:
+    async def get_browsing_history_action(page: Page, _) -> dict[str, Any]:
```
broerjuang: just wonder why we need a second parameter? since action only send page as a callback argument

PR #674, opened 2025-11-19
Remove the sign-in and extract orchestrators

ariya:
This is WIP on how MCP GetGather looks like once any non-distillation code is removed.

Note that the remaining REST API is only for Activities. IMO we should try to move Activities into MCP resources (accessible from MCP Inspector, which we need anyway for day-to-day development).


broerjuang:
Do we also want to remove the frontend components that depend on the deleted API routes? The affected files are Home.tsx, Link.tsx, and Start.tsx. Or should we clean those up in a separate PR?

Screenshot 2025-11-21 at 15 36 48

broerjuang (approved):

PR #681, opened 2025-11-20
Implement distilled page using Zendriver

ariya:
Only BBC tool (to get the saved article) is converted to this Zendriver-based distill page.

To verify, launch MCP Inspector and connect to `/mcp-media` and use the BBC tool `bbc_get_saved_articles`. Continue with the link given for the distilled page to sign in with BBC email and password. After the sign-in is completed, invoking the tool for the second time should display the list of saved articles in JSON.

Because of the minor refactoring, it's also important to check that there is no regression with the Patchright version of distilled page. Hence, launch MCP Inspector to connect to `/mcp-books` and test `goodreads_get_book_list` (it should still work as expected).

### DIFF
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -484,12 +484,11 @@ async def zen_post_dpage(page: zd.Tab, id: str, request: Request) -> HTMLRespons
     if logger.isEnabledFor(logging.DEBUG):
         await zen_capture_page_artifacts(page, identifier=id, prefix="dpage_debug")
 
+    hostname: str | None = getattr(page, "hostname", None)  # type: ignore[assignment]
+
     for iteration in range(max):
         logger.debug(f"Iteration {iteration + 1} of {max}")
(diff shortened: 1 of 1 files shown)


getgather/mcp/dpage.py
```diff
@@ -30,28 +31,44 @@
...
+    run_distillation_loop as zen_run_distillation_loop,
+    terminate_zendriver_browser,
+    zen_report_distill_error,
+)
 
 router = APIRouter(prefix="/dpage", tags=["dpage"])
 
 
-active_pages: dict[str, Page] = {}
+active_pages: dict[str, Page | zd.Tab] = {}
```
ariya: A union like this, so we don't need to keep track of Patchright pages and Zendriver tabs separately, and thus we can use the same `/dpage` route (for the sign-in link).

ariya:
The failing MCP tests are legit, it's not flaky. It exposes some stability issue in the Zendriver version of distilled page, so I'll be working on fixing that.

PR #682, opened 2025-11-21
feat: get ip address tool

faisalive:
Screenshot 2025-11-21 at 11 00 43


getgather/mcp/patterns/ifconfig.html
```diff
@@ -0,0 +1,15 @@
+<html gg-domain="ifconfig.me">
```
ariya: `ifconfig.me` proved to be less reliable than `ip.fly.dev` (which we are using, search for `CHECK_URL`).

faisalive: Sure, will change this

PR #686, opened 2025-11-24
Migrate Goodreads to use Zendriver distillation

ariya:
To verify, use MCP Inspector to invoke `goodreads_get_book_list` tool as usual. Alternatively, try it with Page Turner.


getgather/mcp/patterns/goodreads-modal.html
ariya: Not sure why but I need to add this extra pattern to make sign-in with Goodreads working properly.

Issue #570, opened 2025-10-27
Decommission the sign-in orchestrator


ariya:
Once the critical MCP tools are fully migrated (Wayfair, Nordstrom, etc), we should still remove the sign-in orchestration. If some MCP tools are not converted yet, we will add it step by step in the next cycle, treating it as a new tool.

ariya:
Completed in PR #674.

PR #688, opened 2025-11-24
chore: remove hosted link frontend

broerjuang:
Since #570 has been merged. We can safely remove unused FE that depends on the old sign-in flow


broerjuang:
#689

PR #729, opened 2025-12-02
refactor(aliexpress) zendriver migration

shabrina-ha:
Problem:
Same issue as before when migrating to dpage, when I try it, I encountered email verification code. This is an iframe, so I won't deal with it yet for zendriver. 
If anyone can try it with their account and don't see any email verification code, I think this PR should work. Please help. Thanks!


faisalive:
Success signin, but cannot continue to orders page because we need hover action ([see](https://github.com/remotebrowser/middleman/issues/33))



https://github.com/user-attachments/assets/5e76cf7b-9640-48a3-a5a6-63d5077c850f

PR #742, opened 2025-12-02
Fix ip check order

kpprasa:
There was a subtle bug that prevented us from actually getting the IP when doing the check. Basically we would authenticate with the proxy **after** going to the check url. Thus, when we were routing through oxylabs we would get no IP but the system would work since afterwards we would navigate to a blank page and then go to the location. 

This pr fixes that by standardizing the logic to _always_ go to the blank page first and then all navigations (including ip verification) happen after.

**Before:**
Screenshot 2025-12-02 at 11 10 51 AM

**After**
Screenshot 2025-12-02 at 2 03 22 PM


broerjuang (approved):
nice catch!

PR #743, opened 2025-12-02
Add ip logging

kpprasa:
AT.


getgather/zen_distill.py
```diff
@@ -195,6 +195,21 @@ async def init_zendriver_browser(id: str | None = None) -> zd.Browser:
             # Create page with proxy setup first, then navigate
             page = await get_new_page(browser)
             await page.get(CHECK_URL)
+
+            # Extract and log just the IP address
+            try:
+                await page.wait(2)
```
bin-ario: probably can do `page.expect_response` instead of hardcoded wait

PR #744, opened 2025-12-03
simplify ip logging

kpprasa:
As @ariya mentioned, we can simplify getting the ip by going to the ip endpoint directly.

### DIFF
--- a/getgather/zen_distill.py
+++ b/getgather/zen_distill.py
@@ -186,30 +186,31 @@ async def _create_zendriver_browser(id: str | None = None) -> zd.Browser:
 
 async def init_zendriver_browser(id: str | None = None) -> zd.Browser:
     MAX_ATTEMPTS = 3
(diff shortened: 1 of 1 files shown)


getgather/zen_distill.py
```diff
@@ -186,30 +186,31 @@ async def _create_zendriver_browser(id: str | None = None) -> zd.Browser:
...
             try:
-                await page.wait(2)
-                page_content = await page.get_content()
-                # Extract just IP (format: "ip_address X.X.X.X")
-
-                ip_match = re.search(r"ip_address\s+([\d.]+)", page_content)
-                if ip_match:
-                    logger.info(f"Browser IP address: {ip_match.group(1)}")
+                await ip_page.get(IP_ONLY_CHECK_URL)
+                await ip_page.wait(2)
```
ariya: Does it work without this `wait`? IIRC `get` already performs some basic waiting.

PR #745, opened 2025-12-03
Use a simple static front-end

ariya:
See previously PR #688, #699, and #721.

Most of the front-end plumbing has been simplified. There is no need for OpenAPI schema generation etc.

To verify, run `npm run dev` and open `localhost:23456` as usual. Some other endpoints (`/health`, `/api/docs-mcps`, etc) should still work as well.


broerjuang (approved):
LGTM 🚢 it

PR #753, opened 2025-12-04
change element.click to pointerEvent

faisalive:
somehow `element.click()` not work for some website (eg: https://www.horizonhobby.com/). 

Example: 
1. Open https://www.horizonhobby.com
2. Close popup dialog if present
3. Try using `element.click()`, will not trigger anything
```python
(() => {{
    let element = document.querySelector("div[data-test='account-dropdown'] button[aria-label='Sign In']");
    if (element) { element.click(); return true; }
    return false;
}})()
```
4. Try using pointerEvent, will trigger click
```python
(() => {{
    let element = document.querySelector("div[data-test='account-dropdown'] button[aria-label='Sign In']");
    if (!element) return false;
    element.scrollIntoView({ block: "center" });
    element.dispatchEvent(new PointerEvent("pointerdown", { bubbles: true }));
    element.dispatchEvent(new PointerEvent("pointerup", { bubbles: true }));
    return true;
}})()
```

https://github.com/user-attachments/assets/12e99e42-66ef-4ead-bfb5-fa2f65636213

### DIFF
--- a/getgather/zen_distill.py
+++ b/getgather/zen_distill.py
@@ -375,6 +375,7 @@ class Element:
                 element.scrollIntoView({{ block: "center" }});
                 element.dispatchEvent(new PointerEvent("pointerdown", {{ bubbles: true }}));
                 element.dispatchEvent(new PointerEvent("pointerup", {{ bubbles: true }}));
+                element.dispatchEvent(new MouseEvent("click", {{ bubbles: true, cancelable: true, view: window }}));
                 return true;
             }})()
             """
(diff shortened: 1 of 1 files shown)


faisalive:
> Great idea! Though I think you still need to dispatch a `MouseEvent('click')` at the end. Otherwise, we may risk some compatibility issues.
> 
> Alternatively, as this is also useful to simulate hover events, is to introduce `gg-point` attribute to the distillation pattern. When the attribute exists on an element, then PointerEvents (including `pointerover`) will be dispatched to the matching DOM element.

Sure!, I've added mouse click event

PR #791, opened 2025-12-18
fix: render the page when prev and current distilled still same after timeout

faisalive:
https://heyario.sentry.io/issues/7118213904/events/064567fe258b4df1aacbfc0f0b939783/?project=4509832551858176

Currently, when we receive the same page, the distillation loop continues without any condition. Because of this, if a user receives the same page multiple times (e.g., entering the wrong email repeatedly), the user will eventually see a “Timeout Reached” page.

With this fix, we will still render the distilled page if the page requires user input and the timeout is reached due to receiving the same page again (i.e., the previous and current pages are the same).


broerjuang (commented):
LGTM 🚢

broerjuang (approved):

PR #793, opened 2025-12-18
fix: improve error handling in order details retrieval

oshimayoan:
Will resolve this kind of error where the whole process is stopped because of a single error from a single `page.evaluate()`

https://heyario.sentry.io/issues/7119367965/events/465c49e4825842f4a7b77b85a0191d49/

## How this was tested
Cursor 2025-12-18 16 33 26

## Result
Firefox Developer Edition 2025-12-18 16 35 22


broerjuang:
Can you please do quick regression on Grabbit since order history is using it?

PR #796, opened 2025-12-19
improve log formatting

kpprasa:
This pr tries to systematize what each of Loguru and RichHandler do. Can't totally eliminate RichHandler because the formatting and aesthetics is far better.  
Loguru:
  - Captures log calls
  - Extracts metadata (time, level, location, extras)
  - Formats the entire message including Rich markup tags like
  <level>, <cyan>
  - Passes formatted string to RichHandler

  RichHandler:
  - Receives the already-formatted string from Loguru
  - Only handles Rich markup rendering (colors, styles)
  - Outputs to stderr with pretty tracebacks
  - Does NOT add its own timestamps/levels/paths (all disabled)

### DIFF
--- a/getgather/logs.py
+++ b/getgather/logs.py
@@ -1,6 +1,6 @@
 import logging
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Mapping
 
 import logfire
 import yaml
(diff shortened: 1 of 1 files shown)


kpprasa:
closing for now due to newer logging

PR #834, opened 2025-12-24
feat(distill): gg-error attribute

broerjuang:
The goal of this PR is to:
- [x] Add gg-error to clearly identify whether a detected pattern represents an error (e.g., CAPTCHA).
- [ ] Add proper handling when an error pattern is found (clean up the browser profile and prompt the user to re-sign in). - since we send the pattern already. The client decided what to do

Screenshot 2025-12-25 at 12 18 43

Now the client have extra info about the error.

### DIFF
--- a/getgather/distill.py
+++ b/getgather/distill.py
@@ -379,21 +379,6 @@ async def terminate(distilled: str) -> bool:
     return False
 
 
-def get_stop_error(distilled: str) -> str | None:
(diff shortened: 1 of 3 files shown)


broerjuang:
https://github.com/remotebrowser/mcp-getgather/issues/833

faisalive (approved):

PR #855, opened 2026-01-06
fix amazon continue pattern

faisalive:
- remove gg-match `form[action]` (can have different value)
- change has-text to xpath
- 
image


broerjuang (approved):
LGTM 🚢

PR #896, opened 2026-01-16
feat: proxy configurations

shabrina-ha:


ariya:
@shabrina-ha @kpprasa I think this can be closed for now. It's safe to assume that we will not pursue this path.

PR #907, opened 2026-01-17
Remove MCP tools which rely on the deprecated non-zen dpage

ariya:
Since Zendriver is the future, let's remove the legacy tools!


ariya:
Closing this in favor of #890.

PR #898, opened 2026-01-16
Fix logging

kpprasa:
We were having overly verbose logging due to 3rd party logs being intercepted and captured. 

This should greatly simplify our log setup.

Screenshot 2026-01-15 at 6 23 07 PM
Screenshot 2026-01-15 at 7 44 58 PM

PR #914, opened 2026-01-20
introduce wait for page to be ready

kpprasa:
This fixes a race condition that was exposed by 
- [ ] #898 

where the bbc tool (and maybe others) would consistently fail with on the dpage link. 

During debugging I found that a screenshot helped and that led me to this solution. 

Try it yourself with the BBC tool. 

I am purposely not updating with main for now so the bug will be there if the code isn't there.

Screenshot 2026-01-19 at 5 54 43 PM

### DIFF
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -41,6 +41,7 @@ from getgather.zen_distill import (
     page_query_selector,
     run_distillation_loop as zen_run_distillation_loop,
     safe_close_page,
+    wait_for_ready_state,
     zen_navigate_with_retry,
     zen_report_distill_error,
 )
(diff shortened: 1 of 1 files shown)


getgather/mcp/dpage.py
```diff
@@ -506,12 +507,20 @@ async def zen_post_dpage(page: zd.Tab, id: str, request: Request) -> HTMLRespons
     if settings.LOG_LEVEL == "DEBUG":
         await zen_capture_page_artifacts(page, identifier=id, prefix="dpage_debug")
 
+    # Force browser to complete rendering by evaluating document state
+    try:
+        await wait_for_ready_state(page, timeout=5)
+        # Additional wait for any dynamic content/JavaScript to settle
+        await page.sleep(1)
```
ariya: This is fine for now, but later please check if `sleep(0)` also helps (simply yielding to the main thread).

ariya (approved):

PR #920, opened 2026-01-20
Revert "Revert "Fix logging""

kpprasa:
Reverts remotebrowser/mcp-getgather#911


broerjuang (commented):
wonder why do we revert this back @kpprasa? what's the context. Anyway LGTM, ship it

PR #930, opened 2026-01-21
Remove both Playwright and Patchright as dependencies

ariya:
Note that the CI runs still utilizes Playright (as a tool, not a project dependency) to download Chromium as this proves to be more reliable than installing Chromium natively on GitHub runner (which happens to be using Snap, and not native .deb package).


ariya:
image

PR #931, opened 2026-01-22
clean up configs

bin-ario:
remove unused configs

### DIFF
--- a/getgather/config.py
+++ b/getgather/config.py
@@ -47,12 +47,6 @@ class Settings(BaseSettings):
         return path
 
     @property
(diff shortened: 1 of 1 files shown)


getgather/config.py
```diff
@@ -42,12 +34,6 @@ def data_dir(self) -> Path:
         path.mkdir(parents=True, exist_ok=True)
         return path
 
-    @property
-    def bundles_dir(self) -> Path:
```
ariya: In addition, I think `persistent_store_dir` is also unused.

PR #950, opened 2026-01-28
Working Proxy verification 

kpprasa:
Screenshot 2026-01-27 at 4 17 50 PM

It is very useful to have a self-contained sanity check that the IP configuration actually worked and this is a clean way to do it.


ariya (approved):
Good for now, assuming a forthcoming refactoring.

PR #953, opened 2026-01-28
use AuthUser.user_id as the browser_id

bin-ario:
makes it easy to find an AuthUser's chrome container/app, the names will be like `chrome-123456-google`
if auth is disabled, use a fake user with random user_id, the container name will be like `chrome-xyz123-noauth`

verified with tests/manual.py

### DIFF
--- a/.github/workflows/deploy-fly.yml
+++ b/.github/workflows/deploy-fly.yml
@@ -53,7 +53,7 @@ jobs:
         env:
           FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
         run: |
-          cat .env | flyctl secrets import --app ${{ matrix.org.app_name }}
+          cat .env | flyctl secrets import --app ${{ matrix.org.app_name }} --stage
 
       - name: Deploy ${{ matrix.org.app_name }} app
         env:
--- a/getgather/browser/chromefleet.py
+++ b/getgather/browser/chromefleet.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import Literal
+from typing import Literal, cast
 from urllib.parse import urlparse
 
 import httpx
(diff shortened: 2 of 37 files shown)


getgather/auth/auth.py
```diff
@@ -137,3 +156,12 @@ def get_auth_user() -> AuthUser:
...
     return AuthUser(sub=sub, auth_provider=provider, name=name, email=email, app_name=app_name)
+
+
+NO_AUTH_PROVIDER = "noauth"
+
+
+def _get_user_for_no_auth() -> AuthUser:
+    """Fake auth user for when auth is disabled to keep the code consistent."""
+    sub = generate(FRIENDLY_CHARS, 6)
+    return AuthUser(sub=sub, auth_provider=NO_AUTH_PROVIDER)
```
ariya: @faisalive @broerjuang Please confirm that this new format, e.g. `xyz123-noauth` will not break `signin` etc, for demo apps.

kpprasa: cool so it sounds like the user_id will always avoid the underscores which I think is needed

getgather/auth/auth.py
```diff
@@ -99,14 +101,32 @@ class AuthUser(BaseModel):
...
     @property
     def user_id(self) -> str:
-        """Unique user name combining login and auth provider"""
-        return f"{self.sub}.{self.auth_provider}"
+        """
+        Unique user name combining login and auth provider.
+        Only numbers, lowercase letters and dashes are allowed.
+        Maximum length is 54 characters.
+        """
+        return f"{self.sub}-{self.auth_provider}"
```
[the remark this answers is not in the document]
bin-ario: no strong preference, more like firstname - lastname

PR #955, opened 2026-01-29
Adhere to reference implementation

kpprasa:
Adhering to the chromefleet implementation here: https://github.com/remotebrowser/chromefleet

Tested against this with NPR and it is working
Screenshot 2026-01-29 at 1 54 12 PM

### DIFF
--- a/getgather/browser/chromefleet.py
+++ b/getgather/browser/chromefleet.py
@@ -43,12 +43,7 @@ async def _call_chromefleet_api(method: HTTP_METHOD, browser_id: str) -> httpx.R
     url = f"{base_url}/api/v1/browsers/{browser_id}"
 
     async with httpx.AsyncClient(timeout=30.0) as client:
-        if method.upper() == "POST":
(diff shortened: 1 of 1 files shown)


getgather/browser/chromefleet.py
```diff
@@ -33,24 +35,20 @@ async def _wait_for_cdp(url: str, timeout_s: float = 60.0) -> None:
...
+    url = f"{base_url}/api/v1/browsers/{browser_id}"
+
     async with httpx.AsyncClient(timeout=30.0) as client:
-        response = await client.get(url)
+        if method.upper() == "POST":
+            response = await client.post(url)
+        elif method.upper() == "DELETE":
+            response = await client.delete(url)
+        else:
+            response = await client.get(url)
```
bin-ario: `client.request(method, ...)` works for all cases

kpprasa: oh cool! Didn't know that

PR #956, opened 2026-02-02
feat: garmin tools

oshimayoan:
Firefox Developer Edition 2026-02-02 12 51 36

### DIFF
--- a/getgather/mcp/patterns/garmin-activities.html
+++ b/getgather/mcp/patterns/garmin-activities.html
@@ -12,53 +12,53 @@
         "columns": [
           {
             "name": "activity_icon",
-            "selector": "div.ActivityListItem_activityIconContainer__erO90 i"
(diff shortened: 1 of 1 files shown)


getgather/mcp/patterns/garmin-activities.html
```diff
@@ -0,0 +1,67 @@
...
+        "rows": "div.ActivityListItem_listItem__aoJjV",
+        "columns": [
+          {
+            "name": "activity_icon",
+            "selector": "div.ActivityListItem_activityIconContainer__erO90 i"
+          },
+          {
+            "name": "activity_date",
+            "selector": "span.ActivityListItem_activityDate__iZtT6"
+          },
```
faisalive: Can we use something like this for the selector?

`span[class^="ActivityListItem_activityDate_"]` .

PR #957, opened 2026-02-03
Set location chromefleet

kpprasa:
With chromefleet allowing for location to be passed in addition to a url, this pr safely adapts the code to be more specific when requesting a proxy while still falling back gracefully to the hardcoded url env var. 

Screenshot 2026-02-02 at 5 11 42 PM

### Testing
Easiest way to test it is to run this and point it to a version of chromefleet that supports the location in configure. Then hit it with mcp inspector providing an x-location header


ariya (approved):
LGTM, ship it 🚀

PR #963, opened 2026-02-08
mcp tool support cdp_url via headers

faisalive:


ariya:
@faisalive Is this intended to support OpenClaw use case?
I'm afraid it might not be entirely compatible with the future design, with the interplay with Chrome Fleet. It's probably fine to keep the branch around for illustration purpose.

faisalive:
> @faisalive Is this intended to support OpenClaw use case? I'm afraid it might not be entirely compatible with the future design, with the interplay with Chrome Fleet. It's probably fine to keep the branch around for illustration purpose.

Yes, this only for openclaw use case.

PR #965, opened 2026-02-09
Simplify the termination of a remote browser

ariya:
This will be useful in a forthcoming refactoring, as there is no need to keep the browser id around (the zd.Browser instance is all that is necessary).


bin-ario (approved):

kpprasa:
Is there any reason why it's easier to pass in the browser rather than the browser_id? 

ie `await terminate_remote_browser(browser)` vs `await terminate_remote_browser(browser.id)` ?
Guess it's a preference thing

PR #979, opened 2026-02-11
dont pass empty location

kpprasa:


getgather/request_info.py
```diff
@@ -17,3 +17,8 @@ class RequestInfo(BaseModel):
 
 
 request_info: ContextVar[RequestInfo | None] = ContextVar("request_info", default=None)
+
+
+def is_empty_request_info(info: RequestInfo) -> bool:
+    """Check if all fields in RequestInfo are None."""
+    return all(not getattr(info, field_name) for field_name in RequestInfo.model_fields.keys())
```
bin-ario: would be better to make this a property of RequestInfo

PR #981, opened 2026-02-12
Experimental Goodreads tool which works with Chrome Fleet

ariya:
To try this, first run Chrome Fleet as usual (make sure to have a valid and ephemeral TS_AUTHKEY) and open its web UI.

After that, run this MCP server pointing to the above Chrome Fleet, e.g.

```bash
CHROMEFLEET_URL=http://localhost:8300 npm run dev
```

Then open MCP Jam (or Inspector), connect to the MCP Server, and invoke the tool named `goodreads_remote_get_book_list`. For the first run, the response will be to continue to sign in on a given link. While that happens, the web UI of Chrome Fleet should show a browser container being started and activated.

The dpage link will look like:

```
http://localhost:23456/dpage/xyz123--7CDFB88F40251CC6CE4D50309485CECE
```

In that link, there is the separator `--`. The left side indicates the browser id, and it's also visible on Tailscale as `chromium-xyz123` machine. The right side denotes the tab id from CDP, as also captured by Zendriver. These two components are used to identify where to go (which browser, and which tab) by any subsequent POST handling, and thereby eliminating the need of any housekeeping (like `active_pages` in the case of local Zendriver).

Note that the tab id is rather long for now. Since it's a unique hash, a future tweak will shorten this to the first 8 characters (likely more than sufficient, without any conflict).

2026-02-10 mcp remotebrowser chromefleet goodreads

### DIFF
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -142,9 +142,8 @@ async def get_dpage(id: str | None = None) -> HTMLResponse:
     if id:
         if id in active_pages:
             return redirect(id)
-        else:
-            if is_remote_browser(id):
-                return redirect(id)
+        elif is_remote_browser(id):
+            return redirect(id)
 
     raise HTTPException(status_code=400, detail="Missing page id")
 


getgather/browser/chromefleet.py
```diff
@@ -48,6 +48,32 @@ async def _call_chromefleet_api(method: HTTP_METHOD, browser_id: str) -> httpx.R
...
+    logger.info(f"Finding the ChromeFleet browser: {browser_id}")
+    try:
+        response = await _call_chromefleet_api("GET", browser_id)
+        if response.status_code != 200:
+            return None
+        data = response.json()
+    except Exception:
+        return None
+
+    data = response.json()
```
faisalive: I think we can remove one of `data = response.json()` line

getgather/mcp/dpage.py
```diff
@@ -557,3 +585,62 @@ async def zen_dpage_with_action(
...
+
+    user_id = get_auth_user().user_id
+
+    browser_id: str = user_id
+    browser = await get_remote_browser(browser_id)
+    if browser is None:
+        browser = await create_remote_browser(browser_id)
+
+    page = await get_new_page(browser)
+    dpage_id = f"{browser_id}--{page.target_id}"
```
bin-ario: as an improvement later, this can be modeled as a @dataclass to be less error-prone

getgather/browser/chromefleet.py
ariya: Good catch, thank you! I've removed it accordingly.

PR #986, opened 2026-02-13
Handle incognito mode for remote dpage with Chrome Fleet

ariya:
When `x-incognito` is set, an ephemeral remote browser will be created (marked with `E` prefix in its browser id, e.g. `Exyz123`). This particular instance of remote browser will be used for subsequent distilation operation, linked by its `x-signin-id` (which is now simply mapped to the id of that dpage).

Just like in PR #981, for this remote dpage, the id consists of two parts, the browser id (indexed by Chrome Fleet to find the browser instance) and the page id (corresponding to the actual zd.Tab). That way, there is no need to keep track of its state at all.

This remote browser will be terminated at the end of the distillation process. In the future, Chrome Fleet will have a built-in feature for implicit termination, after some idle time, on an ephemeral instance.

To try this, run this MCP server with Chrome Fleet, per the instructions in PR #981.

After that, launch Page Turner pointing to this MCP server and with the following change (to use the "remote" version of the tool instead). Sign in to Goodreads as usual with Page Turner and at the end, it should show the book list. While that happens, pay attention to Chrome Fleet as it launches a new remote browser (and tears it down, after a while).

```diff
diff --git a/src/server.ts b/src/server.ts
index a8f0b65..53891fe 100644
--- a/src/server.ts
+++ b/src/server.ts
@@ -86,7 +86,7 @@ app.get('/health', (req, res) => {
 app.post('/api/get-book-list', async (req, res) => {
   try {
     const result = await callToolWithReconnect({
-      name: 'goodreads_get_book_list',
+      name: 'goodreads_remote_get_book_list',
       sessionId: req.sessionID,
       ipAddress: getClientIp(req),
     });
```

2026-02-12 page turner remote browser chromefleet

2026-02-12 page turner mcp ephemeral

### DIFF
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -13,7 +13,11 @@ from loguru import logger
 from nanoid import generate
 
 from getgather.auth.auth import get_auth_user
-from getgather.browser.chromefleet import create_remote_browser, get_remote_browser
+from getgather.browser.chromefleet import (
+    create_remote_browser,
+    get_remote_browser,
+    terminate_remote_browser,
+)
 from getgather.config import settings
 from getgather.mcp.browser import browser_manager, terminate_zendriver_browser
 from getgather.mcp.html_renderer import DEFAULT_TITLE, render_form
@@ -47,6 +51,10 @@ pending_actions: dict[str, dict[str, Any]] = {}
 element_configs: dict[str, ElementConfig] = {}
 
 FRIENDLY_CHARS: str = "23456789abcdefghijkmnpqrstuvwxyz"
+
+
+def is_remote_browser(dpage_id: str) -> bool:
+    return "--" in dpage_id
 
 
 async def dpage_add(
@@ -107,6 +115,13 @@ async def dpage_finalize(id: str):
         await terminate_zendriver_browser(browser)
         browser_manager.remove_incognito_browser(id)
         return True
+
+    if is_remote_browser(id):
+        browser_id, _ = id.split("--")
+        if browser := await get_remote_browser(browser_id):
+            await terminate_remote_browser(browser)
+            return True
+
     raise ValueError(f"Browser profile for signin {id} not found in incognito browser profiles")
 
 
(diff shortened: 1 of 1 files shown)


getgather/mcp/dpage.py
```diff
@@ -596,16 +607,42 @@ async def remote_zen_dpage_mcp_tool(
...
+        browser_id, page_id = signin_id.split("--")
+        dpage_id = signin_id
+        browser = await get_remote_browser(browser_id)
+        if browser is None:
+            raise HTTPException(status_code=400, detail="Remote browser not found")
+        for tab in browser.tabs:
+            if tab.target_id == page_id:
+                page = tab
+                break
+        if page is None:
```
ariya: This logic is needed frequently, I will soon refactor it to its own function.

getgather/mcp/dpage.py
```diff
@@ -596,16 +607,42 @@ async def remote_zen_dpage_mcp_tool(
...
+            raise HTTPException(status_code=400, detail="Page not found")
+        logger.info(f"Continue with browser {browser_id} and page {page_id}")
+    elif incognito:
+        prefix = "E"  # for Ephemeral
+        browser_id = prefix + generate(FRIENDLY_CHARS, 7)
```
bin-ario: possibly move this logic to `get_auth_user`, similar to anonymous user or guest user

ariya: Let me try to reconcile that in a follow-up clean-up PR.

Fundamentally, it's the conflict of the following approaches: when the user is **not** authenticated, should they get a **randomized** user id?

If that is the case, then there is no need to special case this "incognito" situation. Incoming MCP clients, such as from the demo apps, will be automatically assigned a fresh but random remote browser.

However, it makes is harder for 2 cases:

(a) local development, since every launch or reload of the MCP server will assign another randomized user id, hence instantiating another fresh remote browser on Chrome Fleet. See #975.

(b) self-hosted MCP server, where the user interacts with e.g. LM Studio with their MCP server. Every redeployment will get a new randomized user id, losing the access to the previous one.

If both cases can be solved while maintaining a randomized user id for non-authenticated user, I think it'll be perfect.

bin-ario: good questions
it would be useful to take a step back and decide how users and chrome-live containers map
It's useful to know if browser_id is just an alias of user_id or a user_id can map to multiple browser_id

ariya: I believe it should be 1:1 mapping between user_id and browser_id.

For non-authenticated users, the mapping remains but the browser is ephemeral (explicit/auto destruct after some time/idle).

Issue #992, opened 2026-02-18
Migrate to remote Zendriver


ariya:
MCP tools should use Zendriver connected to a remote browser (via Chrome Fleet)

Short-lived MCP tools:
* [x] #970

Simple MCP tools (Goodreads etc)
* [x] #981 
* [x] #990 
* [ ] BBC
* [ ] Hardcover
* [ ] Thriftbooks

PR #999, opened 2026-02-20
refactor(thrirftbook): use remote zen 

broerjuang:
#992 

Screenshot 2026-02-20 at 15 49 31

Sign-in works and I’m able to fetch the data from remote. However, after closing the tab, it requires re-signing in again.


ariya (approved):
Ship it 🚀

PR #1004, opened 2026-02-24
feat(dpage): add support for remote_zen_dpage_with_action

broerjuang:
#1005 
This PR adds full remote-browser support to the dpage sign-in flow so actions can run after sign-in when the client uses a remote browser (e.g. via remote browser MCP).

**Remote zen dpage**

- **Signin resume:** Resuming after sign-in now works when the page comes from a remote browser. The flow checks `pending_actions` first, then resolves the page from either local `active_pages` or the remote browser (using `browser_id--page_id`).
- **`remote_zen_dpage_with_action`:** New helper that mirrors `zen_dpage_with_action` but uses remote browsers (`create_remote_browser` / `get_remote_browser`). It tries an existing remote session first, then runs the interactive sign-in flow and stores the pending action so the client can complete sign-in and call back to run the action.

**Goodreads**

- New `get_book_details` tool uses `remote_zen_dpage_with_action` to open a Goodreads book URL and return title, author, rating, and description.

https://github.com/user-attachments/assets/f9846e6c-fc1b-4bb2-ad56-2a573bc76971

Screenshot 2026-02-24 at 15 26 20

### DIFF
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -273,9 +273,7 @@ async def zen_post_dpage(page: zd.Tab, id: str, request: Request) -> HTMLRespons
                 logger.info(f"Signin completed for {id}, resuming action...")
 
                 resume_fn = (
-                    remote_zen_dpage_with_action
-                    if is_remote_browser(id)
-                    else zen_dpage_with_action
+                    remote_zen_dpage_with_action if is_remote_browser(id) else zen_dpage_with_action
                 )
                 action_result = await resume_fn(
                     initial_url=action_info["initial_url"],


getgather/mcp/dpage.py
faisalive: I think we also need to modify these lines to add a condition that determines whether to call `zen_dpage_with_action` or `remote_zen_dpage_with_action`.
To make sure that after finishing the sign-in process, we use the `remote_zen_dpage_with_action` function.

https://github.com/remotebrowser/mcp-getgather/blob/b73f63865f89a91f0c12ebeb5ad198a98ccdb789/getgather/mcp/dpage.py#L274-L281

PR #1088, opened 2026-03-31
feat(amazon, amazonca): remove non remote call

shabrina-ha:
search purchase history:
image

get purchase history:
image

search product:
image

PR #1117, opened 2026-04-08
[DO NOT MERGE] Full remote Zendriver browser

ariya:
This illustrates the state of the codebase once every MCP tool is using remote Zendriver browser. The simplification means that we're down to only ~30 Python source files.

Depends on #1088. ~might need a rebase after #1104~ (already rebased, after that PR was merged).

This is mainly for broader review, not for immediate merging. Likely I will need to create a series of small PRs to ease code review.


ariya:
This big PR was already broken down into smaller PRs and most of them are already merged. Subsequent tweaks will happen in another series of follow-ups.

PR #1120, opened 2026-04-09
Exit on startup if CHROMEFLEET_URL is not set

ariya:


ariya:
@broerjuang Just to confirm, do have any deployed instance where CHROMEFLEET_URL is not set?

bin-ario (approved):

broerjuang:
No I think we don't have any @ariya

PR #1122, opened 2026-04-09
feat: pass x-target-domains to chromefleet

faisalburhanudin:
## Summary

- derive `x-target-domains` from `initial_url`
- send that header when creating a remote browser in Chromefleet

### DIFF
--- a/getgather/browser/chromefleet.py
+++ b/getgather/browser/chromefleet.py
@@ -80,6 +80,7 @@ async def _call_chromefleet_api(
     method: HTTP_METHOD,
     browser_id: str,
     *,
+    initial_url: str | None = None,
     timeout: float = 120.0,
     retries: int = 3,
     raise_for_status: bool = True,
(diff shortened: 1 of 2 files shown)


getgather/browser/chromefleet.py
```diff
@@ -100,6 +100,7 @@ async def _call_chromefleet_api(
         "x-proxy-type": mcp_headers.get("x-proxy-type", None),
         "x-origin-ip": mcp_headers.get("x-origin-ip", None),
         "x-origin-ua": mcp_headers.get("x-origin-ua", None),
+        "x-target-domains": mcp_headers.get("x-target-domains", None),
```
ariya: While this is good as the first ratchet, `x-target-domains` can be set by each MCP tool already (does not have to be from the MCP client/demo apps).

faisalburhanudin: Ah yes, got it. let me rework on it

PR #1127, opened 2026-04-10
feat: forward client IP to ChromeFleet

broerjuang:
Captures the real client IP from MCP requests (Claude Code, Cursor, or the demo web app) and forwards it to ChromeFleet as `x-origin-ip` so proxies can be opened at the correct user location.

**IP resolution priority:**
1. `x-origin-ip` — explicitly set by controlled clients (demo web app)
2. `Fly-Client-IP` — Fly.io deployment
3. `X-Forwarded-For` — GCP / reverse proxy
4. TCP connection IP → if local/private, falls back to `api.ipify.org` to get the machine's real public IP (handles local dev where server and client are on the same machine)

## Verification

**Local (both Cursor and server on same machine):**
1. Start server: `uvicorn getgather.main:app --reload --host 127.0.0.1 --port 23456`
2. Trigger any MCP tool call from Cursor/Claude Code
3. Check logs — should see `[CLIENT IP] <your public IP> (source: server-public-ip)`

**Local server + remote client via ngrok:**
1. Start server: `uvicorn getgather.main:app --host 0.0.0.0 --port 23456`
2. Expose: `ngrok http 23456`
3. Point Cursor MCP config to the ngrok URL
4. Trigger a tool call — logs should show `[CLIENT IP] <real IP> (source: x-forwarded-for)`

Change VPN to Albania 

Screenshot 2026-04-10 at 12 46 13

Screenshot 2026-04-10 at 12 46 38


getgather/main.py
```diff
@@ -218,6 +251,36 @@ async def mcp_logging_context_middleware(
...
+        forwarded_for = request.headers.get("x-forwarded-for")
+        if origin_ip:
+            client_ip = origin_ip
+            ip_source = "x-origin-ip"
+        elif fly_client_ip:
+            client_ip = fly_client_ip
+            ip_source = "fly-client-ip"
+        elif forwarded_for:
+            client_ip = forwarded_for.split(",")[0].strip()
+            ip_source = "x-forwarded-for"
```
ariya: Not sure XFF needs to be honored, as various network elements (routers, load balancer, etc) might inject XFF header when proxying the traffic. It's hard to debug, and we don't want to accidentally adopt an IP address of some random device.

broerjuang: You’re right, this is also confirmed in [Fly.io’s docs](https://fly.io/docs/networking/request-headers/#fly-client-ip). XFF[0] can be spoofed, and Fly-Client-IP only works when Fly is the outermost proxy.  Happy to drop XFF entirely if needed, fallback would be using the server’s public IP (via ipify).

ariya: I think we should ignore both `fly-client-ip` and `x-forwarded-for` for now. If there is a need to add it later, it's always possible.

ariya (changes requested):

getgather/main.py
broerjuang: Sure, will do

ariya (approved):
Ship it 🚀

faisalive (approved):

faisalive:
check are failing @broerjuang

broerjuang:
I'll check it ASAP. Thanks

ariya:
Unfortunately this needs some rebase/merge after the removal of the logging middleware, and also the refactoring in PR #1170.

PR #1128, opened 2026-04-11
perf: reduce span and log noise in dpage_check polling loop

broerjuang:
## Problem

A single `check_signin` request was generating excessive spans and log entries, making logfire traces noisy and hard to read.

## Changes

| File | Change |
|------|--------|
| `dpage.py` | Cache browser object for the duration of the poll; reset to `None` when tab is gone or probe throws so it re-fetches on the next tick |
| `dpage.py` | Load distillation patterns once before the loop and pass them down to `_probe_page`, avoiding redundant disk I/O on every tick |
| `chromefleet.py` | Demote `get_remote_browser` log from INFO to DEBUG to reduce log noise |

## Notes

- `report_error` param for `run_distillation_loop` was merged separately as #1130
- Browser caching fix (tab creation) was merged separately as #1132

### DIFF
--- a/getgather/browser/chromefleet.py
+++ b/getgather/browser/chromefleet.py
@@ -134,7 +134,7 @@ async def _call_chromefleet_api(
 
 
 async def get_remote_browser(browser_id: str) -> zd.Browser | None:
-    logger.info(f"Finding the ChromeFleet browser: {browser_id}")
+    logger.debug(f"Finding the ChromeFleet browser: {browser_id}")
     try:
         await _call_chromefleet_api("GET", browser_id)
     except Exception:
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -24,6 +24,7 @@ from getgather.mcp.html_renderer import DEFAULT_TITLE, render_form
 from getgather.zen_distill import (
     ElementConfig,
     Match,
+    Pattern,
     autoclick as zen_autoclick,
     capture_page_artifacts as zen_capture_page_artifacts,
     check_error,
@@ -116,10 +117,16 @@ async def _try_action_with_probe(
 
 
 async def _probe_page(
-    *, location: str | None = None, page: zd.Tab, browser: zd.Browser, timeout: int = 2
+    *,
+    location: str | None = None,
+    page: zd.Tab,
+    browser: zd.Browser,
+    timeout: int = 2,
+    patterns: list[Pattern] | None = None,
 ) -> bool:
-    path = os.path.join(os.path.dirname(__file__), "patterns", "**/*.html")
-    patterns = load_distillation_patterns(path)
+    if patterns is None:
+        path = os.path.join(os.path.dirname(__file__), "patterns", "**/*.html")
+        patterns = load_distillation_patterns(path)
     terminated, _, _ = await zen_run_distillation_loop(
         location=location,
         patterns=patterns,
@@ -178,6 +185,11 @@ async def dpage_check(id: str):
     browser: zd.Browser | None = None
 
     browser: zd.Browser | None = None
+    probe_patterns = None
+
+    if is_remote:
+        path = os.path.join(os.path.dirname(__file__), "patterns", "**/*.html")
+        probe_patterns = load_distillation_patterns(path)
 
     for iteration in range(max):
         logger.debug(f"Checking dpage {id}: {iteration + 1} of {max}")
@@ -198,16 +210,18 @@ async def dpage_check(id: str):
 
         page = _find_tab(browser, target_id)
         if page is None:
-            browser = None  # tab gone, re-fetch next tick
+            browser = None
             continue
 
         try:
-            terminated = await _probe_page(page=page, browser=browser, timeout=2)
+            terminated = await _probe_page(
+                page=page, browser=browser, timeout=2, patterns=probe_patterns
+            )
             if terminated:
                 return True
         except Exception as e:
             logger.warning(f"Remote probe failed for {id}: {e}")
-            browser = None  # reset on error, re-fetch next tick
+            browser = None
 
     return None
 
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -177,6 +177,8 @@ async def dpage_check(id: str):
     remote_parts = id.split("--", 1) if is_remote else None
     browser: zd.Browser | None = None
 
+    browser: zd.Browser | None = None
+
     for iteration in range(max):
         logger.debug(f"Checking dpage {id}: {iteration + 1} of {max}")
         await asyncio.sleep(TICK)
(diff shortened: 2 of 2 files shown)


getgather/browser/chromefleet.py
```diff
@@ -134,7 +134,7 @@ async def _call_chromefleet_api(
 
 
 async def get_remote_browser(browser_id: str) -> zd.Browser | None:
-    logger.info(f"Finding the ChromeFleet browser: {browser_id}")
+    logger.debug(f"Finding the ChromeFleet browser: {browser_id}")
```
broerjuang, line 137: Demoted from INFO — `get_remote_browser` is a low-level lookup called frequently during polling. INFO was appearing once per second in the trace.

getgather/mcp/dpage.py
```diff
@@ -116,10 +117,16 @@ async def _try_action_with_probe(
 
 
 async def _probe_page(
-    *, location: str | None = None, page: zd.Tab, browser: zd.Browser, timeout: int = 2
+    *,
+    location: str | None = None,
+    page: zd.Tab,
+    browser: zd.Browser,
+    timeout: int = 2,
+    patterns: list[Pattern] | None = None,
```
broerjuang, line 125: Made optional so callers in tight loops can pass pre-loaded patterns instead of reloading from disk on every call.

getgather/mcp/dpage.py
```diff
@@ -116,10 +117,16 @@ async def _try_action_with_probe(
...
 ) -> bool:
-    path = os.path.join(os.path.dirname(__file__), "patterns", "**/*.html")
-    patterns = load_distillation_patterns(path)
+    if patterns is None:
+        path = os.path.join(os.path.dirname(__file__), "patterns", "**/*.html")
```
broerjuang, line 128: `load_distillation_patterns` reads and parses 378 HTML files from disk. With `_probe_page` called every second in `dpage_check`, this was 120 unnecessary I/O + parse cycles per sign-in poll.

getgather/mcp/dpage.py
```diff
@@ -128,6 +135,7 @@ async def _probe_page(
         interactive=False,
         close_page=False,
         page=page,
+        report_error=False,
```
broerjuang, line 138: Probe timeouts are expected — the user is still filling in the sign-in form. Without this, every timeout triggered `zen_report_distill_error`: ERROR log + screenshot + Sentry report, 182 times in a normal sign-in flow.

getgather/mcp/dpage.py
```diff
@@ -175,6 +183,13 @@ async def dpage_check(id: str):
     is_remote = is_remote_browser(id)
     remote_parts = id.split("--", 1) if is_remote else None
 
+    browser: zd.Browser | None = None
+    probe_patterns = None
+
+    if is_remote:
+        path = os.path.join(os.path.dirname(__file__), "patterns", "**/*.html")
+        probe_patterns = load_distillation_patterns(path)
```
broerjuang: Patterns loaded once here and passed down to `_probe_page` for the duration of the poll, avoiding repeated disk reads inside the loop.

getgather/mcp/dpage.py
```diff
@@ -187,20 +202,25 @@ async def dpage_check(id: str):
             continue
 
         browser_id, target_id = remote_parts
-        browser = await get_remote_browser(browser_id)
+        if browser is None:
```
broerjuang, line 201: Previously fetched on every tick. Each `get_remote_browser` call makes an HTTP GET to ChromeFleet, which — via distributed tracing — imports ~9 ChromeFleet-side spans back into the trace (proxy reconfiguration, MaxMind lookup, IP check). A 120-second poll was generating ~1,700 spans from ChromeFleet alone.

getgather/mcp/dpage.py
```diff
@@ -187,20 +202,25 @@ async def dpage_check(id: str):
...
         try:
-            terminated = await _probe_page(page=page, browser=browser, timeout=2)
+            terminated = await _probe_page(
+                page=page, browser=browser, timeout=2, patterns=probe_patterns
+            )
             if terminated:
                 return True
         except Exception as e:
             logger.warning(f"Remote probe failed for {id}: {e}")
+            browser = None
```
broerjuang, line 219: Transient connection error — reset forces a fresh WebSocket connection next tick.

getgather/zen_distill.py
```diff
@@ -1220,6 +1220,7 @@ async def run_distillation_loop(
     interactive: bool = True,
     close_page: bool = True,
     page: zd.Tab | None = None,
+    report_error: bool = True,
```
broerjuang, line 1223: Opt-out flag for callers that use this as a polling probe. Defaults to `true` so all existing callers (main flows) are unaffected.

getgather/zen_distill.py
```diff
@@ -1290,14 +1291,15 @@ async def run_distillation_loop(
...
 
-    await zen_report_distill_error(
-        error=ValueError("No matched pattern found"),
-        page=page,
-        profile_id=browser.id,  # type: ignore[attr-defined]
-        location=location or "",
-        hostname=hostname,
-        iteration=max,
-    )
+    if report_error:
```
broerjuang, line 1294: `zen_report_distill_error` logs an ERROR, captures a screenshot, and reports to Sentry — appropriate for unexpected failures, not for expected probe timeouts.

getgather/zen_distill.py
```diff
@@ -1290,14 +1291,15 @@ async def run_distillation_loop(
...
+        await zen_report_distill_error(
+            error=ValueError("No matched pattern found"),
+            page=page,
+            profile_id=browser.id,  # type: ignore[attr-defined]
+            location=location or "",
+            hostname=hostname,
+            iteration=max,
+        )
```
faisalive, line 1302: This changes will also solve [this](https://heyario.slack.com/archives/C049DGC1D2M/p1774926142766619?thread_ts=1774850634.967919&cid=C049DGC1D2M), Thankyou @broerjuang !

getgather/zen_distill.py
ariya: This is a perfect first ratcher.

For the follow-up improvement (and to avoid [Boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap)), ideally this layer violation is fixed. The distillation function should never know how to report an error. Perhaps the quickest solution is something like a dependency injection, i.e. a `reporter` function is passed an optional parameter, which will be invoked upon error.

getgather/mcp/dpage.py
```diff
@@ -176,6 +183,11 @@ async def dpage_check(id: str):
     is_remote = is_remote_browser(id)
     remote_parts = id.split("--", 1) if is_remote else None
     browser: zd.Browser | None = None
+    probe_patterns = None
+
+    if is_remote:
```
faisalive: I think `is_remote` condition is not needed

faisalive (approved):

PR #1129, opened 2026-04-12
Send x-target-domains from initial_url

faisalburhanudin:
## Summary

When creating a remote browser, this change derives x-target-domains from the initial_url hostname and forwards it to ChromeFleet.

### DIFF
--- a/getgather/browser/chromefleet.py
+++ b/getgather/browser/chromefleet.py
@@ -80,7 +80,7 @@ async def _call_chromefleet_api(
     method: HTTP_METHOD,
     browser_id: str,
     *,
-    initial_url: str | None = None,
+    target_domains: list[str] | None = None,
     timeout: float = 120.0,
     retries: int = 3,
     raise_for_status: bool = True,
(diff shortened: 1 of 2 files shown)


ariya:
I think initial URL is less meaningful (semantically) for invoking the Chrome Fleet API. Perhaps send `target_domains` (optional, default to an empty array) as the function argument? This will become the `x-target-domains` header value.

getgather/browser/chromefleet.py
```diff
@@ -80,6 +80,7 @@ async def _call_chromefleet_api(
     method: HTTP_METHOD,
     browser_id: str,
     *,
+    target_domains: list[str] | None = None,
```
broerjuang: wonder why do we need array for this since the value is `hostname` and the value is only one which is string?

faisalburhanudin: because the task is `x-target-domains` so I assume it plural?

faisalburhanudin: but I understand your concern right now it only send single string, should I make it single value for now?

broerjuang: I ask this because [this function](https://github.com/remotebrowser/mcp-getgather/pull/1129/changes#diff-c5724c0feeaae70325079415f39735174f4aafe2443e9edc4378b8994dc43aceR87) and not sure if this needs to be an array.

faisalburhanudin: sure, lets make it simple for now just single domain, we can always refactor later.

faisalburhanudin: updated

shabrina-ha (approved):

PR #1132, opened 2026-04-13
fix: don't re-create browser instance inside dpage check loop

faisalive:
Currently, in the `dpage_check` loop, we always re-create the browser instance. As a result, it takes a lot of time for the `dpage_check` function to finish (15-20 minutes). This is the memory consumption for 3 hits of `dpage_check`.

Screenshot 2026-04-13 at 14 25 42

After fix:
Screenshot 2026-04-13 at 14 27 19

And somehow this changes also fix this issue: https://heyario.sentry.io/issues/7275157321/?project=4509832551858176&query=is%3Aunresolved&referrer=issue-stream

### DIFF
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -175,6 +175,7 @@ async def dpage_check(id: str):
 
     is_remote = is_remote_browser(id)
     remote_parts = id.split("--", 1) if is_remote else None
+    browser: zd.Browser | None = None
 
     for iteration in range(max):
         logger.debug(f"Checking dpage {id}: {iteration + 1} of {max}")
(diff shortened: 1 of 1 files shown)


getgather/mcp/dpage.py
```diff
@@ -188,7 +189,8 @@ async def dpage_check(id: str):
             continue
 
         browser_id, target_id = remote_parts
-        browser = await get_remote_browser(browser_id)
+        if browser is None:
+            browser = await get_remote_browser(browser_id)
```
broerjuang, line 193: Nice find, This will also kinda reduce the logs on this  #1128

broerjuang (approved):

PR #1143, opened 2026-04-14
improvement: Google multiple 2FA flow

anazaim77:
## Summary

 - `google-signin-choose-method.html` — Replaced auto-click on "Enter your password" with a user-selectable method list. Available sign-in methods (phone tap, Google Authenticator, SMS, recovery email) are shown as clickable buttons. Disabled options (aria-disabled="true") are shown as non-interactive greyed-out buttons. Only options present on the live page are rendered via gg-optional.
 - `zen_distill.py` — Fixed `gg-optional` element handling: unmatched optional elements are now removed from the
 pattern DOM (target.extract()) instead of staying with their fallback text. Previously, both the enabled and
 disabled versions of an option would always render regardless of what was on the page.
 - **New**: google-signin-2fa-authenticator.html — Pattern for the Google Authenticator TOTP code input page.
 - **New**: google-signin-2fa-authenticator-wrong-code.html — Same page after a wrong code is entered (higher
 priority gg-priority="2"), re-prompts for a new code.
 - **New**: google-signin-2fa-wait-recovery-phone.html — Matches the "Google will send a notification to your
 phones" wait page and auto-clicks the Yes button (button[jsname='LgbsSe']).
 - **New**: youtube-channel-subscriptions-empty.html — Handles the empty subscribed channels state. Updated youtube-channel-subscriptions.html to not match this case.

## Demo
https://screenrec.com/share/NF6bx19j5X

image

### DIFF
--- a/getgather/mcp/patterns/google-signin-choose-method.html
+++ b/getgather/mcp/patterns/google-signin-choose-method.html
@@ -3,12 +3,50 @@
     <title>Google Activity</title>
   </head>
   <body>
-    <div gg-match="//span[contains(normalize-space(.), 'Choose how you want to sign in:')]"></div>
-    <div
-      gg-autoclick
-      gg-match="//li//div//div[contains(normalize-space(.), 'Enter your password')]"
+    <div gg-match="//span[@jsname='I74d0c']"></div>
+
+    <!-- Tap Yes on phone or tablet -->
+    <button
+      name="button"
+      type="submit"
+      value="phone-tap-yes"
+      gg-optional
+      gg-match="//div[@jsname='EBHGs'][not(@aria-disabled='true')][@data-challengetype='39'][.//div[@jsname='fmcmS'][not(contains(., 'recovery'))]]"
     >
-      Enter your password
-    </div>
+      Tap Yes on your phone or tablet
+    </button>
+
+    <!-- Google Authenticator -->
+    <button
+      name="button"
+      type="submit"
+      value="authenticator"
+      gg-optional
+      gg-match="//div[@jsname='EBHGs'][not(@aria-disabled='true')][@data-challengetype='6']"
+    >
+      Get a verification code from the Google Authenticator app
+    </button>
+
+    <!-- SMS verification code -->
+    <button
+      name="button"
+      type="submit"
+      value="sms"
+      gg-optional
+      gg-match="//div[@jsname='EBHGs'][not(@aria-disabled='true')][@data-challengetype='9']"
+    >
+      Get a verification code at your phone
+    </button>
+
+    <!-- Tap Yes on recovery email device -->
+    <button
+      name="button"
+      type="submit"
+      value="recovery-email-tap"
+      gg-optional
+      gg-match="//div[@jsname='EBHGs'][not(@aria-disabled='true')][@data-challengetype='39'][.//div[@jsname='fmcmS'][contains(., 'recovery')]]"
+    >
+      Tap Yes on your recovery email device
+    </button>
   </body>
 </html>
--- a/getgather/mcp/patterns/google-signin-try-phone.html
+++ b/getgather/mcp/patterns/google-signin-try-phone.html
@@ -1,4 +1,4 @@
-<html gg-domain="google" gg-priority="2">
+<html gg-domain="google" gg-priority="-2">
   <head>
     <title>Google Activity</title>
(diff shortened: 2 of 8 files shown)


getgather/mcp/patterns/google-signin-try-phone.html
```diff
@@ -1,4 +1,4 @@
-<html gg-domain="google" gg-priority="-2">
+<html gg-domain="google" gg-priority="2">
```
faisalive: I think this was changed purposely here?

https://github.com/remotebrowser/mcp-getgather/pull/1137/changes/cd1e4338a5365f76dacf88b7aff7f593b2aa140b#diff-fbaa382b6b19e2725d2f8285a0e7b7e7e166611cf2dd61f42d6ea8a373d6de3eL1-R1

anazaim77: reverted. I missed this during rebase conflict

getgather/zen_distill.py
```diff
@@ -1038,6 +1038,7 @@ async def distill(
 
             if optional:
                 logger.debug(f"Optional {selector} has no match")
+                target.extract()
```
anazaim77, line 1041: ## Quick Notes

1. **Fallback text won’t show anymore**
`gg-optional` won’t render fallback text if element not found. Should be fine — we’ve been using it as “include if found” anyway.

2. **Parent-child extraction edge case**
If parent is extracted first, child gets detached. We still call `extract()` on it, but it’s safe (no-op in BeautifulSoup).

3. **`match_count` behavior unchanged**
If nothing matches, pattern won’t be included (match_count = 0). Expected 👍

getgather/mcp/patterns/google-signin-choose-method.html
```diff
@@ -3,12 +3,50 @@
...
+      Get a verification code at your phone
+    </button>
+
+    <!-- Tap Yes on recovery email device -->
+    <button
+      name="button"
+      type="submit"
+      value="recovery-email-tap"
+      gg-optional
+      gg-match="//div[@jsname='EBHGs'][not(@aria-disabled='true')][@data-challengetype='39'][.//div[@jsname='fmcmS'][contains(., 'recovery')]]"
```
broerjuang, line 47: The two challengetype 39 variants are told apart by checking if the inner div text contains "recovery". This will break for non-English Google accounts since the display text gets translated. If there is a more stable attribute on the parent `div[jsname='EBHGs']` that distinguishes the two, that would be safer to use.

A few other things related to this file worth checking:

If all four methods come back with aria-disabled="true", every gg-optional button gets removed and the page renders with just the header and no actions. Worth adding some handling for that dead end.

google-signin-2fa.html (not touched in this PR) still matches phone tap via text content. Depending on the account config it might conflict with this flow. Worth checking if that pattern is still needed or should be updated alongside this change.

Same question for google-signin-2fa-wait.html vs the new google-signin-2fa-wait-recovery-phone.html. The anchor text looks different enough but worth a quick sanity check they do not overlap on the same page.

faisalive (approved):

getgather/mcp/patterns/google-signin-choose-method.html
anazaim77: > The two challengetype 39 variants are told apart by checking if the inner div text contains "recovery". This will break for non-English Google accounts since the display text gets translated. If there is a more stable attribute on the parent div[jsname='EBHGs'] that distinguishes the two, that would be safer to use.

still unable to find the unique identifier between those two

> If all four methods come back with aria-disabled="true", every gg-optional button gets removed and the page renders with just the header and no actions. Worth adding some handling for that dead end.

Google 2FA provides 5 verification options and requires at least 2 methods to be enabled. at least, one method must already be active before enabling 2FA

> google-signin-2fa.html (not touched in this PR) still matches phone tap via text content. Depending on the account config it might conflict with this flow. Worth checking if that pattern is still needed or should be updated alongside this change.

I’ve checked this on my end — there’s no conflict in my testing. It consistently shows the 2FA method pattern

> Same question for google-signin-2fa-wait.html vs the new google-signin-2fa-wait-recovery-phone.html. The anchor text looks different enough but worth a quick sanity check they do not overlap on the same page.

yes, I’ve tested both “Tap Yes” flows. I was using my primary device and my recovery phone

PR #1147, opened 2026-04-15
Replace homepage with full-page iframe to Chrome Fleet

ariya:
With remote Zendriver, the live (desktop) view of the local container is gone.

Remove VNC proxy routes and static `index.html`, as well as orphaned VNC frontend, static assets dir, and unused StaticFiles mount.

Part of https://github.com/remotebrowser/mcp-getgather/pull/1117.


broerjuang (approved):
Nice

PR #1148, opened 2026-04-15
feat: inject error code to dpage HTML result

faisalive:
Previously, we can read the error pattern by reading the `check_signin` tool result. However, since [this PR](https://github.com/remotebrowser/mcp-getgather/pull/1095) removed the distillation result from the `check_signin` tool, we need a new approach to detect the error pattern.

In this PR, the error code is embedded inside the dpage HTML result.

### DIFF
--- a/getgather/zen_distill.py
+++ b/getgather/zen_distill.py
@@ -211,13 +211,15 @@ async def terminate(distilled: str) -> bool:
     return False
 
 
-async def check_error(distilled: str) -> bool:
+async def get_error(distilled: str) -> str | None:
     document = BeautifulSoup(distilled, "html.parser")
-    errors = document.find_all(attrs={"gg-error": True})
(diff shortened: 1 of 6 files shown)


getgather/zen_distill.py
```diff
@@ -211,13 +211,15 @@ async def terminate(distilled: str) -> bool:
...
-        logger.info("Found error elements...")
-        return True
-    return False
+    error_element = document.find(attrs={"gg-error": True})
+    if error_element and isinstance(error_element, Tag):
+        error_value = error_element.get("gg-error")
+        logger.info(f"Found error element: {error_value}")
+        if isinstance(error_value, str):
+            return error_value
+    return None
```
broerjuang, line 222: I think this will work fine. So since the value is string we can just do the checking on client side

broerjuang (approved):

PR #1160, opened 2026-04-16
group same session mcp calls into 1 trace

bin-ario:
to make it easy to find logs of a mcp session, we want to group the all logs/spans from multiple mcp calls within 1 session into 1 logfire trace
since 1 span can only appear in 1 trace, the mcp calls won't appear in the client call trace anymore. to address this problem, we create a dummy bridge span under the client call trace, which links to the mcp call span

all changes should be server side only so it works for any mcp clients

- we need to create custom mcp-session-id since fastmcp doesn't generate one in stateless mode. this doesn't violate mcp protocol and the client uses the mcp-session-id the same way as it's designed
- add a tracing middleware to create a root span and reroute downstream spans from caller span to the root span, so all mcp call logs will be shown under the root span
- update logging middleware to set bidirectional span link so that mcp call span and client call span can link to each other

see example below, all mcp calls are under "MCP root" span, the "Span link" on the detail panel links back to the headline-hub request span, and headline-hub request span has a "MCP bridge" span that has a "Span link" to its mcp call span
Screenshot 2026-04-16 at 3 10 40 PM
https://logfire-us.pydantic.dev/getgather/getgather?q=trace_id%3D%276acd379cf11e4108aab4f9c730a0f160%27+and+span_id%3D%276acd379cf11e4108%27&spanId=6acd379cf11e4108&traceId=6acd379cf11e4108aab4f9c730a0f160&env=-clear-&since=2026-04-16T21%3A38%3A48.539367Z&until=2026-04-16T22%3A38%3A48.539367Z

### DIFF
--- a/getgather/main.py
+++ b/getgather/main.py
@@ -1,9 +1,8 @@
 import ast
 import asyncio
-import json
 from contextlib import AsyncExitStack, asynccontextmanager
 from datetime import datetime
-from typing import Any, Awaitable, Callable, Final
+from typing import Awaitable, Callable, Final
 
 from fastapi import FastAPI, Request
 from fastapi.responses import (
@@ -16,9 +15,10 @@ from loguru import logger
 
 from getgather.auth.auth import setup_mcp_auth
 from getgather.config import settings
-from getgather.logs import instrument_fastapi
+from getgather.logs import MCPLoggingContextMiddleware
 from getgather.mcp.dpage import remote_zen_dpage_mcp_tool, router as dpage_router
 from getgather.mcp.main import MCPDoc, create_mcp_apps, mcp_app_docs
+from getgather.tracing import MCPSessionTraceMiddleware, instrument_fastapi
 
 # Create MCP apps once and reuse for lifespan and mounting
 mcp_apps = create_mcp_apps()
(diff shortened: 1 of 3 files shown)


getgather/main.py
```diff
@@ -90,60 +91,11 @@ async def extended_health():
...
-                if body:
-                    body_json: Any = json.loads(body.decode("utf-8"))
-                    if isinstance(body_json, dict):
-                        params: Any = body_json.get("params", {})  # type: ignore[misc]
-                        if isinstance(params, dict):  # type: ignore[arg-type]
-                            signin_id: Any = params.get("signin_id")  # type: ignore[misc]
-                            if signin_id:  # type: ignore[arg-type]
-                                context["signin_id"] = signin_id
-            except Exception:
-                pass
```
bin-ario, line 125: removed 'signin_id' extraction, since mcp-session-id can be used to find all relevant logs

getgather/main.py
```diff
@@ -90,60 +91,11 @@ async def extended_health():
         return PlainTextResponse(content=f"Error: {e}")
 
 
-@app.middleware("http")
-async def mcp_logging_context_middleware(
```
bin-ario, line 94: moved this to `MCPLoggingContextMiddleware` in logs.py

ariya (approved):
LGTM 👍

faisalive (approved):

broerjuang (approved):

PR #1162, opened 2026-04-17
Replace dynamic auto import of MCP tools

ariya:
Previously, MCP modules (amazon, shopee, etc.) were discovered at runtime via a dynamic auto-import mechanism. This makes it harder to add PyInstaller bundling in the near future.

Thus, that dynamic discovery is replaced with explicit static imports. It is not too bad since this is only necessary for MCP tools with custom code (e.g. Amazon), whereas the majority of other MCP tools are automatically loaded via the declarative approach. See previously #1107 and #1114.


faisalive (approved):
🚀

broerjuang (approved):
Nice! LGTM

PR #1141, opened 2026-04-14
fix: amazon get_browsing_history stuck

faisalive:
Fix [this issue](https://logfire-us.pydantic.dev/getgather/getgather?traceId=019d8a5761cdeaa15bb42b1396019db1&spanId=9757d408e96a653e&q=trace_id%3D%27019d8a5761cdeaa15bb42b1396019db1%27+and+span_id%3D%279757d408e96a653e%27&since=2026-04-14T04%3A25%3A06.619430Z&until=2026-04-14T05%3A25%3A06.619430Z), where the code gets stuck on “Waiting for browsing-history API response". It is probably caused by a race condition: the API has already completed on the browser side, but the listener that waits for the API starts listening afterward.

Test result:
Screenshot 2026-04-14 at 14 03 06

PR #1175, opened 2026-04-21
fix: amazon get_browsing_history stuck

broerjuang:
Re-applying fix from #1141 onto the `local-amazon` branch.

Fix where the code gets stuck on "Waiting for browsing-history API response". Caused by a race condition: the API has already completed on the browser side, but the listener that waits for the API starts listening afterward.

Move `page.reload()` inside the `expect_response` context manager so the listener is registered before the reload fires. Applied to both `get_browsing_history_action` occurrences in `amazon.py`.


broerjuang:
Closing in favor of direct cherry-pick onto local-amazon.

PR #1184, opened 2026-04-22
Add PyInstaller configuration for standalone executable

ariya:
Run `npm run package` to build, then check the result:

```bash
$ ls -l dist/
-rwxr-xr-x 1 a a 42629504 Apr 21 21:57 remotebrowser-mcp*
```


broerjuang (approved):
Image
Image


LGTM. Compiled and can be run

PR #1186, opened 2026-04-22
integrate logfire into cdp websocket requests

bin-ario:
override `websockets.connect` to send traceparent in header

see example, flyfleet cdp logs are grouped under mcp trace
Screenshot 2026-04-22 at 3 41 10 PM

### DIFF
--- a/getgather/browser.py
+++ b/getgather/browser.py
@@ -2,17 +2,20 @@ import asyncio
 import json
 import random
 from collections.abc import Awaitable, Callable
+from contextlib import contextmanager
 from dataclasses import dataclass
-from typing import Literal, TypeVar, cast
+from typing import Any, Literal, TypeVar, cast
 from urllib.parse import urlparse
 
 import asyncio_atexit
 import httpx
+import logfire
 import websockets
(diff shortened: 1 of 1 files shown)


getgather/browser.py
```diff
@@ -24,13 +27,38 @@
...
+
+
 async def _create_browser_from_cdp_websocket(
     websocket_url: str, config: Config | None = None
 ) -> zd.Browser:
     parsed = urlparse(websocket_url)
     host = parsed.hostname or "127.0.0.1"
     host = f"[{host.strip('[]')}]" if ":" in host else host  # handle ipv6 addresses
     port = parsed.port or (443 if parsed.scheme in ("wss", "https") else 80)
+    browser_id = parsed.path.rsplit("/", 1)[-1]
```
ariya: Is this defensive enough against a non-conforming or randomized URL path?

bin-ario: we can have `browser_id` passing in, the callers already have it

ariya (approved):
Ship it 🚀

PR #1187, opened 2026-04-23
fix: doordash stuck on network listener

faisalive:
# Replace `tab.expect_response` with full JS `evaluate`

## Summary

Switches the implementation from Playwright-style `tab.expect_response` to driving the same behavior via full JavaScript `evaluate` in the page context.

## Motivation

`expect_response` was unreliable in practice (race timing, missed or duplicate handlers, or responses that never matched the predicate). Running the logic in-page with `evaluate` aligns execution with the tab’s real JS environment and stabilizes the flow.

## Results (measured)

| Approach           | Success | Error | Success rate |
| ------------------ | ------- | ----- | ------------ |
| `expect_response`  | 6       | 4     | **60%**      |
| `evaluate`         | 10      | 0     | **100%**     |

### DIFF
--- a/getgather/mcp/doordash.py
+++ b/getgather/mcp/doordash.py
@@ -1,4 +1,3 @@
-import json
 from typing import Any, cast
 
 import zendriver as zd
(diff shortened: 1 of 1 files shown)


ariya (approved):
Ship it 🚀

PR #1190, opened 2026-04-23
fix: error google multi 2FA method looping

anazaim77:
## Overview
  - Update Google sign-in pattern matching for alternative authentication steps and errors.
  - Replace/remove outdated selectors and add wrong-password/tap-yes handling to make sign-in flows more reliable.

## Testing
image


faisalive:
- Handle wrong password case
- Fix multi signin method pattern
- Remove duplicate pattern

broerjuang (approved):

PR #1192, opened 2026-04-23
group /dpage logs into mcp session trace

bin-ario:
add a dataclass structure to signin_id, which includes browser_id, target_id and mcp_session_id
mcp_session_id is added so we can group /dpage logs into the same mcp trace
rename dpage_id to signin_id for consistency
Screenshot 2026-04-23 at 12 50 54 PM


broerjuang (approved):
LGTM

PR #1193, opened 2026-04-23
Use an optional error reporter callback in the main distillation loop

ariya:
This is refactoring only, no behavior change at all. It's a follow-up improvement to PR #1130.

Replace the `report_error: bool` parameter with an optional `error_reporter` callback, improving the semantics of the core distillation loop. Callers now explicitly opt in to error reporting by passing a reporter function, rather than relying on a boolean flag.

Add `make_error_reporter()` helper that wraps `zen_report_distill_error` with browser/location context, and an `ErrorReporter` type alias.

### DIFF
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -36,6 +36,7 @@ from getgather.zen_distill import (
     get_error,
     get_selector,
     load_distillation_patterns,
+    make_error_reporter,
     run_distillation_loop as zen_run_distillation_loop,
     terminate,
     zen_report_distill_error,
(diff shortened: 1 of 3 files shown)


getgather/mcp/dpage.py
```diff
@@ -131,7 +132,6 @@ async def _probe_page(
         interactive=False,
         close_page=False,
         page=page,
-        report_error=False,
```
ariya, line 134: Avoiding the anti-pattern "Boolean trap": https://ariya.io/2011/08/hall-of-api-shame-boolean-trap

bin-ario (approved):

broerjuang (approved):

PR #1209, opened 2026-04-28
feat(google): add email-not-found sign-in pattern

anazaim77:
## Overview
Added new patter to handle email not found after submit an email

## Testing
https://screenrec.com/share/QWXuSGtUA5

### DIFF
--- a/getgather/mcp/patterns/google-signin-email-notfound.html
+++ b/getgather/mcp/patterns/google-signin-email-notfound.html
@@ -5,6 +5,6 @@
       gg-match="//div[@jsname='B34EJ'][contains(., 'find your Google Account')]"
       style="color: red; font-size: 12px"
     ></div>
-    <button type="submit" gg-autoclick gg-match="div#identifierNext">Next</button>
+    <button type="submit" gg-match="div#identifierNext">Next</button>
   </body>
 </html>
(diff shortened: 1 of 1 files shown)


getgather/mcp/patterns/google-signin-email-notfound.html
```diff
@@ -0,0 +1,10 @@
+<html gg-domain="google" gg-priority="-2">
+  <body>
+    <input autofocus placeholder="Email" gg-match="input#identifierId" type="email" name="email" />
+    <div
+      gg-match="//div[@jsname='B34EJ'][contains(., 'find your Google Account')]"
```
ariya, line 5: Just be careful with this likely-generated symbol `B34EJ`, as it may not be the same in the future.

getgather/mcp/patterns/google-signin-email-notfound.html
```diff
@@ -0,0 +1,10 @@
...
+      style="color: red; font-size: 12px"
+    ></div>
+    <button type="submit" gg-autoclick gg-match="div#identifierNext">Next</button>
```
faisalive: For the submit button element, you can use either `type="submit"` or `gg-autoclick`, there’s no need to use both.

getgather/mcp/patterns/google-signin-email-notfound.html
anazaim77: yes, I’m aware of that. so far, `jsname` looks more stable compared to class names but let’s keep an eye on it

getgather/mcp/patterns/google-signin-email-notfound.html
ariya: And in that case, stick with `button[type=submit]` (semantically more correct).

anazaim77: image

it is happening everywhere 🫠 will open a PR to handle this

ariya (approved):

PR #1211, opened 2026-04-29
Avoid the explicit browser create

ariya:
Since Chrome Fleet auto-starts on first CDP, drop the explicit call `POST /api/v1/browsers`.

However, the proper headers still need to be correctly populated.

- Extract `_build_chromefleet_headers()` helper shared by HTTP and websocket.
- Inject MCP headers (X-Forwarded-For, Client Hints, X-Origin-*) into the CDP websocket handshake via `_inject_traceparent_into_websockets`.

See also https://github.com/remotebrowser/chromefleet/pull/48.

### DIFF
--- a/getgather/browser.py
+++ b/getgather/browser.py
@@ -27,8 +27,24 @@ from getgather.config import settings
 HTTP_METHOD = Literal["GET", "POST", "DELETE"]
 
 
+def _build_chromefleet_headers(*, target_domain: str | None = None) -> dict[str, str]:
+    mcp_headers = get_http_headers(include_all=True)
+    headers = {
+        "x-forwarded-for": mcp_headers.get("x-forwarded-for", None),
+        "user-agent": mcp_headers.get("user-agent", None),
+        "sec-ch-ua": mcp_headers.get("sec-ch-ua", None),
+        "sec-ch-ua-mobile": mcp_headers.get("sec-ch-ua-mobile", None),
+        "sec-ch-ua-platform": mcp_headers.get("sec-ch-ua-platform", None),
+        "x-origin-ip": mcp_headers.get("x-origin-ip") or client_ip_var.get(),
+        "x-origin-id": mcp_headers.get("x-origin-id", None),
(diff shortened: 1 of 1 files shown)


getgather/browser.py
```diff
@@ -207,19 +215,18 @@ async def get_remote_browser(browser_id: str) -> zd.Browser | None:
...
 ) -> zd.Browser:
     """
-    Start a remote Chrome via ChromeFleet and connect via CDP.
-    The browser_id must not already be in use.
+    Connect to a remote Chrome via ChromeFleet CDP.
+    ChromeFleet auto-starts the browser on first CDP access if it doesn't exist.
     """
-    logger.info(f"Starting new ChromeFleet browser: {browser_id}")
-    await _call_chromefleet_api("POST", browser_id, target_domain=target_domain)
+    logger.info(f"Connecting to ChromeFleet browser: {browser_id}")
```
faisalive, line 224: Does FlyFleet also support auto-start on the first CDP call?

getgather/browser.py
```diff
@@ -27,8 +27,24 @@
...
+        "x-origin-id": mcp_headers.get("x-origin-id", None),
+        "x-origin-ua": mcp_headers.get("x-origin-ua", None),
+        "x-target-domains": target_domain,
+    }
+    return {k: v for k, v in headers.items() if v is not None}
+
+
 @contextmanager
-def _inject_traceparent_into_websockets():
+def _inject_traceparent_into_websockets(extra_headers: dict[str, str] | None = None):
```
bin-ario: It would make sense to rename this function to `_inject_headers_into_websockets`

bin-ario (approved):

getgather/browser.py
ariya: Yes, that feature has been added.

faisalive (approved):

getgather/browser.py
ariya: Good point! The function has been renamed accordingly.

broerjuang (approved):

PR #1213, opened 2026-04-29
fix: remove redundant gg-autoclick from submit controls

anazaim77:
## Issue
https://github.com/remotebrowser/mcp/pull/1209#discussion_r3155096306

## Summary
- Removes redundant `gg-autoclick` from pattern elements that already use `type="submit"` across 24 HTML pattern files.
- Keeps behavior explicit and unambiguous by relying on type="submit" as the canonical submit trigger in those flows.
- Fixes one special-case pattern by removing meaningless type="submit" from a <div> while preserving gg-autoclick in getgather/mcp/patterns/amazon-keep-hacker-out.html.

## Why
- Using both type="submit" and gg-autoclick on the same element is redundant and can imply conflicting/double-submit behavior.
- This cleanup improves readability and makes intent clearer for future pattern maintenance.

### DIFF
--- a/getgather/mcp/patterns/adidas-password.html
+++ b/getgather/mcp/patterns/adidas-password.html
@@ -12,7 +12,7 @@
       gg-match="input#password"
       placeholder="Password"
     />
-    <button type="submit" gg-autoclick gg-match="button#login-submit-button:not([disabled]) span">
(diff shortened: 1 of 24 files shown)


ariya (approved):
Ship it 🚀

broerjuang (approved):
Nice catch!

PR #1214, opened 2026-04-29
migrate github workflows to shared github actions

bin-ario:
Migrate `container.yml` and `deploy-fly.yml` to shared actions https://github.com/remotebrowser/shared-github-actions
similar to https://github.com/gather-engineering/flyfleet/pull/131


fly.toml
```diff
@@ -3,7 +3,6 @@
 # See https://fly.io/docs/reference/configuration/ for information about how to use this file.
 #
 
-app = 'remotebrowser-prod'
```
broerjuang, line 6: is it intended?

broerjuang (approved):
LGTM

fly.toml
bin-ario: yes, we don't have an app named 'remotebrowser-prod'
and the script runs `fly deploy --app <name>` so we can deploy the code to multiple apps

ariya (approved):

PR #1216, opened 2026-04-30
CI to test executable packaging with PyInstaller

ariya:


ariya:
This is mainly to prevent any regressions (like the one fixed in PR #1215).

bin-ario (approved):

PR #1219, opened 2026-04-30
fix: isolate CDP websocket headers per request

faisalive:
## Summary
- replace nested per-call websockets.connect monkey-patching with a single global wrapper plus ContextVar-scoped extra headers
- prevent concurrent requests from overriding each other's x-target-domains during CDP websocket handshake
- add logger.info diagnostics for context set/reset and effective handshake headers

## Test plan
- [x] ReadLints clean for getgather/browser.py
- [ ] Run concurrent flows (e.g. amazon + youtube with x-incognito) and verify CDP logs show correct per-request x-target-domains
- [ ] Confirm no regressions in regular (non-incognito) sign-in flows

### DIFF
--- a/getgather/browser.py
+++ b/getgather/browser.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import random
+from contextvars import ContextVar
 from collections.abc import Awaitable, Callable
 from contextlib import contextmanager
 from dataclasses import dataclass
@@ -25,6 +26,11 @@ from getgather.client_ip import client_ip_var
 from getgather.config import settings
 
 HTTP_METHOD = Literal["GET", "POST", "DELETE"]
+_ws_extra_headers_var: ContextVar[dict[str, str] | None] = ContextVar(
+    "_ws_extra_headers_var", default=None
+)
+_ws_connect_patched = False
+_original_websockets_connect: Any = None
 
 
 def _build_chromefleet_headers(*, target_domain: str | None = None) -> dict[str, str]:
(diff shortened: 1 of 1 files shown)


getgather/browser.py
```diff
@@ -46,27 +80,21 @@ def _build_chromefleet_headers(*, target_domain: str | None = None) -> dict[str,
...
+    # Install a single process-wide wrapper once, then pass per-request headers
+    # via ContextVar so concurrent requests do not overwrite each other.
+    _ensure_ws_connect_patched()
+
+    token = _ws_extra_headers_var.set(extra_headers or None)
+    logger.info(
+        "Set websocket header context: target_domain={}, header_keys={}",
+        (extra_headers or {}).get("x-target-domains"),
+        sorted((extra_headers or {}).keys()),
+    )
```
broerjuang: I don't think we need to log all the headers.

getgather/browser.py
```diff
@@ -105,6 +133,11 @@ async def _safe_handle_target_update(event: object) -> None:
             logger.debug("Ignored transient target update race: StopIteration")
 
     extra_headers = _build_chromefleet_headers(target_domain=target_domain)
+    logger.info(
+        "Opening CDP websocket for browser_id={} with target_domain={}",
+        browser_id,
+        extra_headers.get("x-target-domains"),
+    )
```
broerjuang: this too?

getgather/browser.py
faisalive: removed

getgather/browser.py
faisalive: removed

broerjuang (approved):
:shipit:

PR #1221, opened 2026-04-30
Experiment with built-in browser container management

ariya:
This is mainly for development, to avoid the need to run Chrome Fleet separately. It's not yet well tested, though the code flow is designed to avoid separate branches for built-in vs remote fleet functionality.

To verify, run as usual with `npm run dev` without the need to specify `CHROMEFLEET_URL` env and the MCP server should still work as expected.


getgather/residential_proxy.py
bin-ario, line 1: is the proxy code needed? didn't see it used anywhere

ariya: It's set from `getgather/browsers_api_router.py`.

However, if you think the first ratchet should not include any proxy code, that's also fair (and I can remove it).

bin-ario (approved):

getgather/residential_proxy.py
bin-ario: Works for me as is

PR #1226, opened 2026-05-04
feat: block traffic for specific path/file via Zendriver

anazaim77:


ariya:
(same question) What's an example of this path-level blocking optimization?

PR #1238, opened 2026-05-07
feat: error pattern logs

faisalive:
Next phase of #1231: We want to differentiate between the success pattern and the error (gg-optional) pattern. To do this, we also need to add the error message to Logfire.

Screenshot 2026-05-07 at 14 18 39

### DIFF
--- a/getgather/zen_distill.py
+++ b/getgather/zen_distill.py
@@ -328,7 +328,7 @@ def make_error_reporter(browser: zd.Browser, location: str | None = None) -> Err
 async def distill(
     hostname: str | None, page: zd.Tab, patterns: list[Pattern], reload_on_error: bool = True
 ) -> Match | None:
-    result: list[Match] = []
+    result: list[tuple[Match, list[str]]] = []
     pattern_runs: list[tuple[str, int, BeautifulSoup, list[dict[str, object]]]] = []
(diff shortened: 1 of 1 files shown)


getgather/zen_distill.py
```diff
@@ -418,6 +419,23 @@ async def distill(
...
+                        if isinstance(raw_value, str):
+                            text_content = raw_value.strip()
+                    if not text_content:
+                        raw_html = source.get("html")
+                        if isinstance(raw_html, str):
+                            text_content = BeautifulSoup(raw_html, "html.parser").get_text(
+                                " ", strip=True
+                            )
+                    if text_content:
+                        optional_visible_text.append(text_content)
```
oshimayoan, line 438: I think we can simplify it like this:

```py
if optional:
    text_content = ""
    for key in ("text", "value"):
        raw = source.get(key)
        if isinstance(raw, str) and raw.strip():
            text_content = raw.strip()
            break
    if not text_content:
        raw_html = source.get("html")
        if isinstance(raw_html, str):
            text_content = BeautifulSoup(raw_html, "html.parser").get_text(" ", strip=True)
    if text_content:
        optional_visible_text.append(text_content)
```

Wdyt?

oshimayoan (approved):

PR #1244, opened 2026-05-12
feat: handle page not found pattern

faisalburhanudin:
https://github.com/gather-engineering/demos/issues/739

### DIFF
--- a/getgather/mcp/patterns/amazon-error-page-not-found.html
+++ b/getgather/mcp/patterns/amazon-error-page-not-found.html
@@ -1,0 +1,14 @@
+<html gg-domain="amazon">
+  <head>
+    <title>Amazon Page Not Found</title>
+  </head>
+  <body>
+    <div
+      gg-stop
+      gg-error="page_not_found"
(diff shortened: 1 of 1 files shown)


faisalive:
I think we can find a workaround to handle this pattern. This appears randomly. Sometimes, when we retry the login flow on the same browser with the same account, it successfully opens the Amazon orders page.

faisalburhanudin:
Got it, close it for now

getgather/mcp/patterns/amazon-error-page-not-found.html
```diff
@@ -0,0 +1,14 @@
+<html gg-domain="amazon">
+  <head>
+    <title>Amazon Page Not Found</title>
+  </head>
+  <body>
+    <div
+      gg-stop
+      gg-error="captcha"
+      gg-match="//div[@id='g']//p[contains(normalize-space(.), 'find that page')]"
+    >
```
faisalive, line 10: This error can be solved by retrying the connection. So, making the behavior similar to a “captcha” flow (re-creating the browser) will likely solve the issue.

faisalive (approved):

PR #1250, opened 2026-05-13
feat: handle dynamic challenge credit card

faisalburhanudin:
https://github.com/gather-engineering/demos/issues/739

## Summary

Handle patten dynamic challenge question credit card

## Note

need to develop select


faisalive:
We can use `gg-error="reset_password"` for this, so on the Frontend side, it will show specific error page that tell user to try their account on amazon.com

PR #1262, opened 2026-05-20
feat: init walmart tool

oshimayoan:
- Adds `walmart_get_order_history` MCP tool that extracts purchase history from the Walmart `/orders` page
- Implements full sign-in distillation flow: email/phone input, password, OTP, and two post-password verification paths (verify method selection and phone-only auto-send)
- Order history is read from the SSR'd `__NEXT_DATA__` script tag (`props.pageProps.phRedesignInitialData.data.purchaseHistory`) rather than intercepting GraphQL, because Walmart SSRs the page on direct URL navigation and the `PurchaseHistoryV3` GraphQL call only fires during client-side SPA navigation

Firefox Developer Edition 2026-05-20 12 29 06

### DIFF
--- a/getgather/mcp/walmart.py
+++ b/getgather/mcp/walmart.py
@@ -1,0 +1,34 @@
+from typing import Any, cast
+
+import zendriver as zd
+
+from getgather.mcp.dpage import remote_zen_dpage_with_action
+from getgather.mcp.registry import MCPTool
+
+walmart_mcp = MCPTool.registry["walmart"]
+
+
+async def _get_order_history_action(tab: zd.Tab, _: zd.Browser) -> dict[str, Any]:
+    result = await tab.evaluate(
+        """
+        (() => {
+            const el = document.getElementById('__NEXT_DATA__');
+            if (!el) throw new Error('__NEXT_DATA__ not found');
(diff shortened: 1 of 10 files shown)


faisalive:
Should we support pagination for this?

oshimayoan:
@faisalive Thinking of adding it on the follow-up diff, for now want to init the tool first

faisalive (approved):

getgather/mcp/walmart.py
```diff
@@ -0,0 +1,34 @@
...
+            if (!el) throw new Error('__NEXT_DATA__ not found');
+            const data = JSON.parse(el.textContent);
+            const purchaseHistory = data?.props?.pageProps?.phRedesignInitialData?.data?.purchaseHistory;
+            if (!purchaseHistory) throw new Error('purchaseHistory not found in __NEXT_DATA__');
+            return purchaseHistory;
+        })()
+        """,
+        await_promise=False,
+    )
+    return {"walmart_order_history": cast(dict[str, Any], result)}
```
broerjuang: I think we should just use `order_hitsory` just like https://github.com/remotebrowser/mcp/blob/main/getgather/mcp/amazon.py#L41

PR #1263, opened 2026-05-21
Update env template settings

faisalive:
## Summary
- Update `.env.template` to match current `Settings` and `AuthSettings` environment variables.
- Remove stale `HOSTNAME` entry and align defaults for `LOG_LEVEL`, ChromeFleet, auth, proxy, GeoIP, and Logfire settings.

## Test plan
- [x] `npm run frontend:check:format -- .env.template`


.env.template
```diff
@@ -1,11 +1,33 @@
...
+# Application settings
 ENVIRONMENT=local
-LOG_LEVEL=DEBUG
+GIT_REV=
+PORT=23456
+DATA_DIR=
 
-# Logging
-SENTRY_DSN=
+# Chrome Fleet
```
ariya: I think we can exclude this entire Chrome Fleet block since it is purely deployment specific and it does not affect any self-hosting folks.

.env.template
```diff
@@ -1,11 +1,33 @@
-# Application Settings
+# Application settings
 ENVIRONMENT=local
-LOG_LEVEL=DEBUG
+GIT_REV=
```
ariya: Same reason, `GIT_REV` is set automatically (for a specific type development), does not affect regular users.

ariya (approved):

PR #1267, opened 2026-05-22
fix: auto-select SMS OTP on Walmart verify method screen

oshimayoan:


oshimayoan:
Closing this as we move to focus on email OTP approach in #1268

PR #1276, opened 2026-05-26
Run distillation tests against built wheel, not Chrome Fleet

ariya:
Build the wheel and self-host Remote Browser directly in CI, removing the dependency on the separate Chrome Fleet app. This ensures distillation tests run against the same binary that gets published to PyPI (continues #1235).

image

### DIFF
--- a/.github/workflows/distill-tests.yml
+++ b/.github/workflows/distill-tests.yml
@@ -11,20 +11,28 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 15
     steps:
+      - name: Prepare the browser container image
+        run: podman pull ghcr.io/remotebrowser/chromium-live
+
       - name: Checkout code
         uses: actions/checkout@v6
-
-      - name: Run Chrome Fleet
-        uses: ./.github/actions/run-chromefleet
-        timeout-minutes: 5
 
       - name: Prepare back-end environment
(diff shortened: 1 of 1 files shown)


.github/workflows/distill-tests.yml
```diff
@@ -11,20 +11,28 @@ jobs:
...
+
+      - name: Run the health check validation
+        run: while ! curl -s 'http://localhost:23456/health' | grep 'OK'; do sleep 1; done
+        timeout-minutes: 3
+
       - name: Run the distillation tests
         run: uv run pytest -m "distill" -s -p no:xdist
         env:
-          CHROMEFLEET_URL: http://localhost:8300
+          CHROMEFLEET_CDP_OPEN_TIMEOUT_SECONDS: 120
```
ariya, line 35: Increased timeout, vs the default, since GitHub Actions machine is slower.

.github/workflows/distill-tests.yml
```diff
@@ -11,20 +11,28 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 15
     steps:
+      - name: Prepare the browser container image
+        run: podman pull ghcr.io/remotebrowser/chromium-live
```
ariya, line 15: This is to get the container image pre-emptively.

bin-ario (approved):

PR #1289, opened 2026-06-01
feat(target): add Target.com sign-in entry point

anazaim77:
## Summary

- Register Target brand in `getgather/mcp/mcp-tools.yaml` and shopping bundle
- Add Target dpage sign-in flow patterns: pre-signin, username, choose-method, password, signin-error
- Add stub `target_get_purchases` tool wired through dpage sign-in flow
- Add delayed autoclick support and `data-testid` updates for Target patterns

## Notes

- Purchase extraction is still stubbed
- Scope is Target sign-in foundation only

## Demos

- Initial sign-in error handled via 20s retry: https://screenrec.com/share/81sGtPZvRd
- Retried flow successful: https://screenrec.com/share/7nulJQ6wvC

### DIFF
--- a/getgather/mcp/patterns/target-order-error.html
+++ b/getgather/mcp/patterns/target-order-error.html
@@ -1,0 +1,14 @@
+<html gg-domain="target" gg-priority="-2" gg-error="order_error">
+  <head>
+    <title>Target Order Error</title>
+  </head>
+  <body>
+    <div gg-match=".styles_ndsAlertCard__YtwCK.styles_warning__u2sW1"></div>
+    <button
+      gg-match="//button[@type='submit' and contains(normalize-space(.), 'Sign in with password')]"
+      gg-autoclick-delay="20"
+    >
+      Sign in with password
+    </button>
+  </body>
+</html>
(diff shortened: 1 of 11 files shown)


getgather/mcp/patterns/target-order-error.html
```diff
@@ -0,0 +1,14 @@
...
+    <title>Target Order Error</title>
+  </head>
+  <body>
+    <div gg-match=".styles_ndsAlertCard__YtwCK.styles_warning__u2sW1"></div>
+    <button
+      gg-match="//button[@type='submit' and contains(normalize-space(.), 'Sign in with password')]"
+      data-testid="signin-method-password"
+    >
+      Sign in with password
+    </button>
```
faisalive: Since there's only one button, can we just use gg-autoclick?

getgather/mcp/patterns/target-order-error.html
```diff
@@ -0,0 +1,14 @@
+<html gg-domain="target" gg-priority="-2" gg-error="order_error" gg-action-delay="20">
```
oshimayoan: Nit: I don't think the action delay would helpful here as there will be only a single pending action from this pattern

getgather/mcp/patterns/target-order-error.html
anazaim77: We’ll stop the user on the error page, so `gg-autoclick` will not be needed.

getgather/mcp/patterns/target-order-error.html
anazaim77: Reverted, as we don’t need the delay config for this PR

faisalive (approved):

PR #1291, opened 2026-06-02
fix: add delays between input focus, fill, and submit actions

oshimayoan:
During Walmart signin, the submit button was firing before the email field was fully filled, a potential race condition that caused the demo app to ask email again in the followup form.

- Add gg-action-delay attribute on pattern <body> to opt into delays per brand
- When set, delays are applied between each batch action (fill, submit click) and after focusing an input before typing starts
- Walmart signin email pattern opts in with gg-action-delay="200"
- All other brands are unaffected (no attribute = no delay)

### DIFF
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -554,7 +554,7 @@ async def zen_post_dpage(page: zd.Tab, id: str, request: Request) -> HTMLRespons
             results = action_results if isinstance(action_results, dict) else {}
 
             # Fallback for failed/unexecuted actions to preserve behavior.
-            for action in pending_actions:
+            for i, action in enumerate(pending_actions):
                 key = action.get("key")
                 kind = action.get("kind")
                 selector = action.get("selector")
@@ -567,6 +567,8 @@ async def zen_post_dpage(page: zd.Tab, id: str, request: Request) -> HTMLRespons
                 if results.get(key, False):
                     continue
 
+                if i > 0:
+                    await asyncio.sleep(0.3)
                 element = await page_query_selector(page, selector)
                 if not element:
                     continue
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -362,8 +362,8 @@ async def zen_post_dpage(page: zd.Tab, id: str, request: Request) -> HTMLRespons
         options = {"title": title, "action": action}
         inputs = document.find_all("input")
         pending_actions: list[dict[str, str]] = []
-        body = document.find("body")
-        action_delay_ms = int(str(body.get("gg-action-delay") or 0)) if isinstance(body, Tag) else 0
+        html_element = document.find("html")
+        action_delay_ms = int(str(html_element.get("gg-action-delay") or 0)) if isinstance(html_element, Tag) else 0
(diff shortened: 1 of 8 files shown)


getgather/mcp/dpage.py
```diff
@@ -567,14 +571,19 @@ async def zen_post_dpage(page: zd.Tab, id: str, request: Request) -> HTMLRespons
                 if results.get(key, False):
                     continue
 
+                delay_ms = int(action.get("action_delay_ms") or 0)
+                if i > 0 and delay_ms > 0:
+                    await asyncio.sleep(delay_ms / 1000)
                 element = await page_query_selector(page, selector)
```
faisalive: why do we need to check `i > 0`?

getgather/mcp/dpage.py
```diff
@@ -567,14 +571,19 @@ async def zen_post_dpage(page: zd.Tab, id: str, request: Request) -> HTMLRespons
...
                 if not element:
                     continue
                 if kind == "click":
                     await element.click()
                 elif kind == "set_value":
                     value = action.get("value")
-                    await element.type_text(value if isinstance(value, str) else "")
+                    await element.type_text(
+                        value if isinstance(value, str) else "", focus_delay_ms=delay_ms
+                    )
```
faisalive: There are already `asyncio.sleep` on line 575, why we need to also do it inside `type_text`?

getgather/mcp/dpage.py
oshimayoan: To make sure there is more than single action on the queue, because the idea is to add delay between action

getgather/mcp/dpage.py
oshimayoan: Oops, you're right. Fixed.

ariya:
For another page, @faisalive added `class ElementConfig` some time ago. Can this Walmart MCP tool use a similar approach, e.g. using the configuration in that ElementConfig class instead>

faisalive:
> For another page, @faisalive added `class ElementConfig` some time ago. Can this Walmart MCP tool use a similar approach, e.g. using the configuration in that ElementConfig class instead>

Currently, we're not using ElementConfig. I think that's because we migrated it to Zendriver.

oshimayoan:
@faisalive From what I understand, based on the code, ElementConfig is still used as the fallback when the JS in browser.py fails. Perhaps that is what you mean by not using it anymore?

@ariya I think we can indeed use it for the python implementation. Here is the changes [035aab9](https://github.com/remotebrowser/mcp/pull/1291/commits/035aab94483446b40af2b16c7251046e2228c2b4)

faisalive:
> @faisalive From what I understand, based on the code, ElementConfig is still used as the fallback when the JS in browser.py fails. Perhaps that is what you mean by not using it anymore?
> 
> @ariya I think we can indeed use it for the python implementation. Here is the changes [035aab9](https://github.com/remotebrowser/mcp/pull/1291/commits/035aab94483446b40af2b16c7251046e2228c2b4)


Yes, the code still there, but I think there are no pattern that define `element_config` attribute.
Screenshot 2026-06-04 at 22 08 21

And also dpage tool function not using that variable.
Screenshot 2026-06-04 at 22 08 49

faisalive (approved):

PR #1294, opened 2026-06-03
fix: remove browser profile dirs on terminate and sweep orphaned profiles

broerjuang:
## Problem

Since ~01:00 UTC on Jun 3, tap-amazon production logged **477 \"Failed to start browser after 3 attempts\"** errors (~470 distinct browsers in 3 hours). Root cause: the machine disk filled up with browser profile dirs under `/app/data/profiles`, so Chrome could no longer start.

Profile dirs were never deleted:

- `terminate_zendriver_browser` only pruned cache subdirs (GPUCache, Code Cache, etc.) — the profile itself (Cookies, Local Storage, IndexedDB, …) stayed on disk forever.
- Browser tracking is in-memory only, so profiles orphaned by crashes/restarts/failed starts were never seen by the cleanup loop.
- Failed `zd.start` attempts and failed browser validations also leaked partial profile dirs.

Profiles are never reopened after termination (`init_zendriver_browser` only looks up in-memory browsers), so all of this was dead weight.

## Changes

- **`terminate_zendriver_browser`** now removes the whole profile dir instead of just cache subdirs.
- **`BrowserManager.cleanup_orphaned_profiles()`** (new): sweeps `profiles_dir` and removes dirs that aren't an active browser and haven't been modified within `BROWSER_SESSION_AGE`. Wired into the existing 5-minute timer in `main.py` via `asyncio.to_thread`.
- **`_create_zendriver_browser`**: removes the partially-initialized profile dir when all start attempts fail.
- **`init_zendriver_browser`**: validation failures now use `terminate_zendriver_browser` (instead of bare `browser.stop()`) so throwaway retry profiles are deleted — including on the final attempt, which previously leaked a running browser.

## Tests

- 10 new tests in `tests/mcp/test_browser_cleanup.py` (full-dir removal, `ConnectionClosed` handling, stale orphan removal, active/global browser protection, recent-dir grace period, stray files, `BROWSER_SESSION_AGE` setting).
- pyright + ruff clean.

## Note for ops

This only stops future growth — currently-full machines still need `/app/data/profiles` cleared (fly ssh or machine restart) to recover.

### DIFF
--- a/getgather/main.py
+++ b/getgather/main.py
@@ -51,6 +51,11 @@ async def lifespan(app: FastAPI):
             except Exception as e:
                 logger.error(f"Error in cleanup_incognito_browsers: {e}", exc_info=True)
             try:
+                # Run in a thread: rmtree of large profile dirs can block for seconds
+                await asyncio.to_thread(browser_manager.cleanup_orphaned_profiles)
+            except Exception as e:
+                logger.error(f"Error in cleanup_orphaned_profiles: {e}", exc_info=True)
+            try:
                 await asyncio.wait_for(stop_event.wait(), timeout=5 * 60)
             except asyncio.TimeoutError:
                 pass  # Timeout = 5 minutes passed, continue loop
--- a/getgather/mcp/browser.py
+++ b/getgather/mcp/browser.py
@@ -1,12 +1,23 @@
+import shutil
 from datetime import datetime, timedelta
+from pathlib import Path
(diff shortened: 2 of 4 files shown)


ariya (approved):
Ship it 🚀

faisalburhanudin (approved):

faisalive (approved):

PR #1295, opened 2026-06-04
feat(walmart): bot detection pattern

oshimayoan:
Adds a distillation pattern for the PerimeterX "Robot or human?" challenge modal that appears on walmart.com. Signals `captcha` error and stops the flow when `#px-captcha` is detected.

### DIFF
--- a/getgather/mcp/patterns/walmart-bot-detection.html
+++ b/getgather/mcp/patterns/walmart-bot-detection.html
@@ -1,0 +1,8 @@
+<html gg-domain="walmart.com" gg-priority="9">
+  <head>
+    <title>Walmart Bot Detection</title>
+  </head>
+  <body>
+    <div gg-stop gg-error="captcha" gg-match="div#px-captcha"></div>
+  </body>
+</html>


getgather/mcp/patterns/walmart-bot-detection.html
```diff
@@ -0,0 +1,8 @@
+<html gg-domain="walmart.com" gg-priority="9">
+  <head>
+    <title>Walmart Bot Detection</title>
+  </head>
+  <body>
+    <div gg-stop gg-error="captcha" gg-match="div#px-captcha"></div>
```
faisalive, line 6: Do we want to handle Walmart bot detection exactly the same way as Amazon captcha detection (i.e., recreate the browser up to 3 times)?

On Amazon, the captcha is shown after the user enters their email address (before signing in).

On Walmart, however, bot detection can sometimes appear after the user signs in and accesses the Orders page.

oshimayoan: Yeah, Walmart is kinda tricky because of that. I don't think following Amazon approach is a good call because of that condition where it could happen after user signed in. We can't confuse user by asking them to re-login right after they just logged in, in case the bot detection shows up in orders page.

But on the other side, I'm not sure whether user will get the bot detection as well when they open Walmart directly on their browser after it happened on flyfleet

broerjuang: How often the bot detection appears on flyfleet compared to lambda?

faisalive (approved):

PR #1297, opened 2026-06-05
feat(walmart): support "Text me" button variant in choose-method pattern

oshimayoan:
Adds "Text me" to the submit button XPath in the choose-method pattern to cover a Walmart sign-in variant that uses that label instead of "Sign in" / "Request code" / "Continue".

Firefox Developer Edition 2026-06-05 16 07 56


faisalive (approved):
🚀

PR #1308, opened 2026-06-08
fix(walmart): apply action delay before first set_value to stop email clearing

oshimayoan:
## Problem
On `identity.walmart.com` the email sign-in field intermittently clears before submit. `set_value` types the email char-by-char, the value is confirmed in the DOM, a few chars flash, then the field resets to empty. Submit fires empty and Walmart returns "Please enter a valid phone number or email". Intermittent on remote FlyFleet.

## Root cause
The input is a React controlled component still mid-render when typing begins, so the re-render wipes the typed value.

## Fix
Apply `action_delay_ms` before the first action (the `set_value`), not just subsequent ones, so React finishes rendering before typing.

- `browser.py` — drop the `index > 0` guard on the batch-loop delay
- `dpage.py` — fallback always uses `element_config`; simplify the now-unused `enumerate`
- `walmart-signin-email.html` — bump `gg-action-delay` 200ms → 1000ms

### DIFF
--- a/getgather/browser.py
+++ b/getgather/browser.py
@@ -984,7 +984,7 @@ async def page_batch_actions(page: zd.Tab, actions: list[dict[str, str]]) -> dic
 
         for (const [index, action] of actions.entries()) {{
             const actionDelayMs = Number(action?.action_delay_ms) || 0;
-            if (index > 0 && actionDelayMs > 0) {{
+            if (actionDelayMs > 0) {{
                 await new Promise(resolve => setTimeout(resolve, actionDelayMs));
             }}
             const key = action?.key;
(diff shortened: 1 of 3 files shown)


oshimayoan:
This is followup from #1291

faisalive:
@oshimayoan can you provide a video record of before and after this changes?

oshimayoan:
@faisalive It's weird that it doesn't work anymore when I try to record it. I've been tinkering with this since morning, will put this diff into draft for now

faisalive:
> @faisalive It's weird that it doesn't work anymore when I try to record it. I've been tinkering with this since morning, will put this diff into draft for now

What you mean are the issue are gone **or** this PR doesn't solve the issue? @oshimayoan

oshimayoan:
The issue persists @faisalive

oshimayoan:
@faisalive Just realized that I was running the wrong worktree, no wonder the issue persists. 🤣

## Before

https://github.com/user-attachments/assets/9dadba68-daaa-46b7-a89b-4091e9f69772

## After

https://github.com/user-attachments/assets/5f616c19-a17a-47bf-927a-43ed0d6d717f

oshimayoan:
Closing this as we'll go with #1328

PR #1311, opened 2026-06-08
Support both stand-alone and external Chrome Fleet operations

ariya:
Remote Browser can operate in two modes: stand-alone or with an external Chrome Fleet.

In stand-alone mode, Podman manages browser containers locally. All browser CRUD endpoints (create, delete, get, list) use the Podman-based container handling.

With an external Chrome Fleet, those same CRUD endpoints are proxied to the upstream Chrome Fleet service.

To verify, first run Chrome Fleet (or any compatible implementation) on port 8300. Then launch Remote Browser as usual, specifying the URL of that Chrome Fleet, e.g.

```bash
export CHROMEFLEET_URL=http://localhost:8300
LOG_LEVEL=DEBUG npm run dev
```

Then invoke the Remote Browser REST API, e.g.

```bash
curl localhost:23456/api/v1/browsers
```

The log of Chrome Fleet should show something like:
```
13:43:52 INFO     Enumerating all browsers...
         INFO     127.0.0.1:41704 - "GET /api/v1/browsers HTTP/1.1" 200
```

while the log of Remote Browser would indicate the proxying of that request to Chrome Fleet:
```
13:43:52 INFO     Proxying list_browsers() to external Chrome Fleet
         INFO     127.0.0.1:33326 - "GET /api/v1/browsers HTTP/1.1" 200
```

### DIFF
--- a/getgather/browsers_api_router.py
+++ b/getgather/browsers_api_router.py
@@ -33,6 +33,26 @@ from getgather.zen_distill import convert, distill, load_distillation_patterns
 router = APIRouter()
 
 
+async def _chromefleet_request(
+    method: str,
+    path: str,
+    *,
+    headers: dict[str, str] | None = None,
+) -> httpx.Response:
+    url = f"{settings.CHROMEFLEET_URL}{path}"
+    async with httpx.AsyncClient(timeout=30.0) as client:
+        return await client.request(method, url, headers=headers)
+
+
+def _forward_headers(request: HTTPConnection, *names: str) -> dict[str, str]:
+    headers: dict[str, str] = {}
+    for name in names:
+        value = request.headers.get(name)
+        if value:
+            headers[name] = value
+    return headers
+
+
 async def get_cdp_websocket_url(browser_id: str) -> str:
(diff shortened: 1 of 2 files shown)


getgather/browsers_api_router.py
```diff
@@ -33,6 +33,26 @@
...
+        return await client.request(method, url, headers=headers)
+
+
+def _forward_headers(request: HTTPConnection, *names: str) -> dict[str, str]:
+    headers: dict[str, str] = {}
+    for name in names:
+        value = request.headers.get(name)
+        if value:
+            headers[name] = value
+    return headers
```
bin-ario: these seem to be duplicates of `_call_chromefleet_api` and `_build_chromefleet_headers` in `browser.py`?

ariya: It's not very similar because the headers are originated from different sources (MCP headers vs regular HTTP headers), but I've added the commit which refactors this and now it should be less duplication.

bin-ario (approved):

PR #1312, opened 2026-06-09
feat: unify browser backends under getgather/browsers with a Daytona backend

bin-ario:
This PR groups all browser-backend management into a new `getgather/browsers` package behind a shared `Backend` protocol, and adds a Daytona backend alongside the existing podman and external-fleet paths. It builds on #1310 (podman extraction) and #1311 (external Chrome Fleet proxying), unifying their podman/fleet split under one factory.

## What changed

A `Backend` protocol (`browsers/backend.py`) that the REST endpoints and CDP/VNC proxies depend on, selected at startup by `create_backend()`:

- `FleetBackend` (`fleet_browsers.py`): proxies the browser API to an external Chrome Fleet when `CHROMEFLEET_URL` is set. `call_chromefleet_api` and `build_chromefleet_headers` move here from `browser.py`, which now imports them back and is the zendriver interface only.
- `DaytonaBackend` (`daytona_browsers.py`): on-demand Daytona sandboxes (`BROWSER_BACKEND=daytona`), reaching Chrome over a signed HTTPS preview URL. Added as an optional `daytona` extra so podman installs are unchanged. Teardown is owned by Daytona's native `auto_stop_interval` + `auto_delete_interval` (per gather-engineering/flyfleet#159).
- `PodmanBackend` (`podman_browsers.py`): the existing podman logic (#1310) wrapped as a backend; stays the default.

`router.py` (renamed from `browsers_api_router.py`) and `residential_proxy.py` move into the package. The router delegates to the selected backend uniformly, replacing the per-endpoint `CHROMEFLEET_URL` branching from #1311. All browser/backend settings live in `BrowserSettings` (`browsers/settings.py`), which `config.Settings` inherits. The CDP proxy gains `rewrite_ws_url` (a no-op for podman) so the same proxy works for the Daytona signed URL.

## Notes for reviewers

- VNC (`/live`, `/websockify`) and the residential-proxy / geo-IP features remain podman only; `DaytonaBackend.get_vnc_endpoint` returns `None` and `FleetBackend` serves CDP via the external fleet's own `/cdp`.
- The local backend selector is `BROWSER_BACKEND` (`podman` default, `daytona`); `CHROMEFLEET_URL` takes precedence over it.
- The Daytona SDK's async client pulls `httpx-ws`, which needs `anyio>=4.10` but under-declares it, so the `daytona` extra pins `anyio>=4.10`.
- The new `test-daytona.yml` workflow needs CI config to run: secret `DAYTONA_API_KEY`, secrets `TS_OAUTH_CLIENT_ID`/`TS_OAUTH_SECRET`, and vars `DAYTONA_API_URL`/`DAYTONA_SNAPSHOT`. It self-skips the Tailscale step on non-github-hosted runners.

## Test plan

- pyright (strict), ruff check + format, yamlfix, prettier all pass.
- Podman (default): create/get/list/404/delete cycle against a running server with identical response shapes.
- Fleet: `FleetBackend` pointed at a live upstream proxied create/list/get/delete and mapped a missing browser to 404.
- Daytona (live, self-hosted): `tests/test_daytona.py` passes, including the native-lifecycle-interval assertions.
- Unit suite passes; the Daytona tests skip under the default backend.


ariya (approved):
Looks great, and pretty clean as well 🚀

.env.template
```diff
@@ -12,11 +12,25 @@ OAUTH_GOOGLE_CLIENT_SECRET=
...
+DAYTONA_API_KEY=
+DAYTONA_API_URL=
+DAYTONA_SNAPSHOT=
+
+# Residential proxy (Massive) — podman backend only
 MASSIVE_PROXY_USERNAME=
 MASSIVE_PROXY_PASSWORD=
 
-# MaxMind GeoIP
+# MaxMind GeoIP — podman backend only
```
ariya: Is there a reason the two above settings are for Podman only?

ariya (approved):

.env.template
bin-ario: residential proxy is not integrated in daytona browsers yet, it will be added soon

bin-ario: Reworded the comment to explain the reason: these are consumed only by PodmanBackend.configure_remote_browser (per-browser upstream proxy + geo from x-origin-ip). Daytona ignores origin_ip, and in fleet mode the upstream Chrome Fleet applies its own proxy/geo. (9b4101e)

PR #1316, opened 2026-06-09
test(patterns): add data-testid to dpage pattern inputs and buttons

bin-ario:
This PR gives every visible `<input>` and `<button>` in `getgather/mcp/patterns` a `data-testid` from a shared vocabulary so automated playwright tests can share locator code across brands. The attributes are additive: distillation returns the pattern skeleton as-is and dpage renders its body verbatim, so nothing existing changes behavior.

The vocabulary:

| data-testid               | Used for                                                       |
| ------------------------- | -------------------------------------------------------------- |
| `email`                   | fields that accept an email address only                       |
| `username`                | combo identifier fields (email or phone/username/member number) |
| `password`                | password fields                                                |
| `phone`                   | phone number fields                                            |
| `otp`                     | single verification-code fields                                |
| `otp-0` … `otp-N`         | multi-box OTP fields, 0-based                                  |
| `submit`                  | the primary submit/continue button                             |
| `signin-method-*`         | sign-in method radios and alternate-method buttons             |
| `option-<n>`              | generic choice radios (e.g. YouTube channel picker)            |
| `zip-code`, `answer`      | Amazon ZIP and security-question fields                        |

For `email` vs `username`, the upstream `gg-match` selector is the source of truth since pattern placeholders can be inaccurate: `input#ap_email` → `email`, `input#enterUsername` / `input#identifierId` / `input[name='loginKey']` → `username`.

Hidden helpers (`display: none`, `type=hidden`), `gg-autoclick` elements, and empty match-only buttons are exempt. Traveloka's pre-existing `gg-*` testids are renamed to the shared vocabulary, including its inline script selectors.

A new script, `scripts/check_pattern_testids.py`, enforces the convention: it fails on missing or unknown `data-testid` values, type mismatches, and non-contiguous or 1-based OTP box ids. It runs in `check:all` (and the pre-push hook) as `patterns:check:testids`, and as a `pattern-testids` job in the static-analysis workflow.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -60,3 +60,16 @@ jobs:
 
       - name: Check YAML formatting
         run: uv run yamlfix --check $(find . -type f -name '*.yml' -o -name '*.yaml' | grep -v -E '\.venv/|node_modules/|mcp-tools\.yaml')
+
+  pattern-testids:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 5
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Prepare back-end environment
+        uses: ./.github/actions/prepare-backend
+
+      - name: Check pattern data-testid convention
+        run: uv run python scripts/check_pattern_testids.py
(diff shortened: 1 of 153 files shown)


broerjuang (approved):
LGTM

PR #1317, opened 2026-06-09
fix(health): use an ephemeral browser for the extended-health probe

bin-ario:
This PR makes `/extended-health` create a short-lived, random-id browser per probe and terminate it when done, instead of reusing the shared noauth browser.

The probe previously ran in the persistent noauth browser, whose id derives from the machine hostname. That caused two problems:

- GitHub-hosted runner hostnames repeat across runs, so a Container-workflow run could inherit the previous run's auto-stopped Daytona sandbox and fail its CDP probe with `400 Bad Request` until auto-delete fired (~75 minutes). This failed the `daytona` job twice on [#1316](https://github.com/remotebrowser/mcp/pull/1316).
- In noauth deployments the hostname-derived browser is the same persistent browser real user flows use, so a health probe shared state with live sessions.

The probe now goes through `short_lived_mcp_tool` (the same primitive the one-off scrape tools use), which gets a `timeout` parameter to keep the probe's previous 3s distillation budget and a `finally` so the browser is terminated even when the probe fails. Each probe pays a browser cold start, which also means it exercises browser creation end-to-end on every run.

Verified against the self-hosted Daytona server: `curl /extended-health` returns `OK IP: ...` in ~4s and the sandbox list is empty afterward.

Note the reconnect-to-stopped-sandbox bug in the Daytona backend (`get_cdp_base_url` not restarting stopped sandboxes) still exists for persistent user browsers and is tracked separately.


getgather/main.py
```diff
@@ -80,8 +81,15 @@ def health():
 @app.get("/extended-health")
 async def extended_health():
     try:
-        result = await remote_zen_dpage_mcp_tool(
-            initial_url="https://ip.fly.dev/ip", result_key="ip_address", timeout=3
+        # A fresh short-lived browser per probe, terminated when done: the probe never touches
+        # the shared noauth browser (which holds real sessions in noauth deployments) and leaves
+        # no sandbox behind for a later run on the same hostname to inherit.
+        _, result = await short_lived_mcp_tool(
```
ariya, line 87: I think eventually `short_lived_mcp_tool` will go away, but this is fine for now.

ariya (approved):

PR #1318, opened 2026-06-10
feat: extract qty from amazon

faisalive:
Extract qty from amazon order detail (normal, fresh, wholefood)

[amazon-with-qty.json](https://github.com/user-attachments/files/28782309/amazon-with-qty.json)


oshimayoan (approved):
LGTM

broerjuang (approved):

PR #1319, opened 2026-06-10
CI: Only run Daytona browser backend tests when DAYTONA_API_KEY is set

ariya:
Add a setup job that computes the backend matrix dynamically and feed it to both E2E test jobs via fromJSON, so the Daytona choice is skipped when the secret is absent.

This way, when someone forks this repo and works on it, and they don't have Daytona account, the CI workflows won't fail on them.

### DIFF
--- a/.github/workflows/browser-api-e2e-test.yml
+++ b/.github/workflows/browser-api-e2e-test.yml
@@ -13,13 +13,33 @@ on:
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


bin-ario (approved):
there is one more workflow that uses DAYTONA_API_KEY
https://github.com/remotebrowser/mcp/blob/main/.github/workflows/container.yml#L83

PR #1322, opened 2026-06-11
Add patterns CRUD API

ariya:
Add `/api/v1/patterns` endpoints (list, get, upsert, delete) backed by the existing patterns directory, with name validation to prevent path traversal. Update README with API documentation.

Add a CBC distillation E2E test that uploads custom patterns (`.html` and `.json` converter) before running extraction, demonstrating the full upload-then-distill workflow for sites without built-in patterns.

image
image

### DIFF
--- a/getgather/patterns_api_router.py
+++ b/getgather/patterns_api_router.py
@@ -1,0 +1,108 @@
+import os
+import re
+
+from fastapi import APIRouter, HTTPException, Query
+from fastapi.responses import JSONResponse, PlainTextResponse
+from loguru import logger
+from pydantic import BaseModel
+
+router = APIRouter()
+
+PATTERNS_DIR = os.path.join(os.path.dirname(__file__), "mcp", "patterns")
+
+_VALID_NAME = re.compile(r"^[a-zA-Z0-9_-]+$")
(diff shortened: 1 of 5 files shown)


bin-ario (approved):

broerjuang (approved):

getgather/patterns_api_router.py
```diff
@@ -0,0 +1,108 @@
...
+async def upsert_pattern(
+    pattern_name: str, body: PatternBody, ext: str = Query("html")
+) -> dict[str, str]:
+    _validate_name(pattern_name)
+    suffix = _validate_ext(ext)
+    path = _pattern_path(pattern_name, suffix)
+    exists = os.path.isfile(path)
+    try:
+        with open(path, "w", encoding="utf-8") as f:
+            f.write(body.content)
```
broerjuang, line 84: Just wondering, do we plan to introduce persistence for patterns in the future, since they appear to be stored on the local filesystem and would be lost on container restarts or redeployments?

ariya: Might be too complex for now. The easiest is for the consumer (e.g. Headline Hub) to upload the pattern before distillation. It's an extra overhead, but it might be good enough for now.

PR #1328, opened 2026-06-11
fix: stabilize input fill against React remount on Walmart sign-in

oshimayoan:
## Problem
Walmart remounts its React component tree during hydration, wiping DOM values set before the remount. The email input appeared filled but was cleared before submit.

## Solution
`set_value` now polls `element.value` every 50ms and re-fills if wiped. It waits for 5 consecutive stable ticks (~250ms) before resolving, so the submit only fires once the value has held.

## Flow
svgviewer-output(8)

## Demo

https://github.com/user-attachments/assets/e7bf0532-ccfe-46e9-947e-697416d231ec

### DIFF
--- a/getgather/browser.py
+++ b/getgather/browser.py
@@ -961,82 +961,38 @@ async def page_batch_actions(page: zd.Tab, actions: list[dict[str, str]]) -> dic
             el.dispatchEvent(new Event("change", {{ bubbles: true }}));
         }}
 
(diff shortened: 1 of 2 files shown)


faisalive:
Do we still need this ?
https://github.com/remotebrowser/mcp/pull/1308

oshimayoan:
@faisalive Nope, that will be closed when this land

faisalive (approved):

PR #1344, opened 2026-06-17
fix: use Daytona signed preview iframe for live view

faisalive:
# Daytona Live View in Iframe

## Summary
- Use Daytona signed preview URL (port `8080`) for live view.
- Show Daytona live view inside `/live/{browser_id}` using an iframe.
- Keep podman live view unchanged (`rfb.min.js` + `/websockify/{browser_id}`).

## Why
- Avoid Daytona preview auth errors in the dashboard.
- Keep the main app page on our domain while still showing live desktop.
- Keep backend behavior clear per provider.

## Changes
- Add `get_live_view_url()` to browser backend interface.
- Implement Daytona `get_live_view_url()` using `create_signed_preview_url(8080)`.
- Return `None` for fleet/podman in `get_live_view_url()`.
- Update `/live/{browser_id}` route:
  - Daytona: render iframe with signed preview URL.
  - Podman: keep existing local noVNC flow.

## Test Plan
- [x] `uv run ruff check getgather/browsers/router.py getgather/browsers/daytona_browsers.py`
- [x] `uv run pyright getgather/browsers/daytona_browsers.py`
- [x] Open `http://localhost:23456/` with `BROWSER_BACKEND=daytona` and verify live views load.

## Test Result

Screenshot 2026-06-17 at 15 40 22
Screenshot 2026-06-17 at 15 40 13

### DIFF
--- a/getgather/browsers/router.py
+++ b/getgather/browsers/router.py
@@ -1,4 +1,5 @@
 import asyncio
+import html
 import json
 import os
 import urllib.parse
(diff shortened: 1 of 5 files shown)


getgather/browsers/router.py
```diff
@@ -507,9 +508,31 @@ async def cdp_devtools_websocket_proxy(client_ws: WebSocket, path: str) -> None:
     logger.debug("[CDP] cdp_devtools_websocket_proxy exiting")
 
 
-@router.get("/live/{browser_id}")
+@router.get("/live/{browser_id}", response_model=None)
 async def vnc_live_viewer(browser_id: str) -> HTMLResponse:
-    html = f"""<!DOCTYPE html>
+    try:
+        live_url = await backend.get_live_view_url(browser_id)
```
bin-ario, line 514: this live url approach could work, however, we should add some gates to prevent the live from keeping the sandbox running unnecessarily for a long time 

for example, we check the sandbox' last activity time, and only return the url if the sandbox is refresh, e.g., within 1 hour

faisalive: sure!

anazaim77:
tested on my local

image

anazaim77 (approved):

PR #1348, opened 2026-06-17
Add periodic cleanup of idle browser (Podman) containers

ariya:
To verify, run as usual and use its API to create a test browser. Wait for some time and the idle cleanup will kick in:

```
         INFO     Application startup complete.
15:21:15 DEBUG    Retrieving the list of all containers...
         DEBUG    All containers obtained. Total=1
         DEBUG    Browser abc001 idle for 101m
         INFO     Deleting browser abc001 (idle: 101m)
         INFO     Killing Chromium container chromium-abc001...
         INFO     Container killed: name=chromium-abc001
         INFO     Cleanup complete: total=1 deleted=1
```

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -19,6 +19,7 @@ BROWSER_BACKEND=podman
 # Podman backend
 CONTAINER_IMAGE=ghcr.io/remotebrowser/chromium-live
 CONTAINER_HOST=
+MAX_IDLE_MINUTES=15
 
 # Daytona backend (required when BROWSER_BACKEND=daytona; install the extra: uv sync --extra daytona).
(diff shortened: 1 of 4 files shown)


.env.template
```diff
@@ -19,6 +19,7 @@ BROWSER_BACKEND=podman
 # Podman backend
 CONTAINER_IMAGE=ghcr.io/remotebrowser/chromium-live
 CONTAINER_HOST=
+MAX_IDLE_MINUTES=15
```
ariya, line 22: This is now configurable via env, in case it's needed.

bin-ario (approved):

PR #1350, opened 2026-06-19
fix(browsers): run all podman subprocesses off the event loop

ariya:
Every async helper in `podman_browsers.py` called `run_podman()` directly inside an `async def`. `run_podman()` uses `subprocess.run()`, which is blocking — so each call held the asyncio event loop for the full duration of the podman invocation.

This made endpoints backed by these helpers (notably `/api/v1/browsers`, but also create/delete/get-browser and the idle cleanup loop) feel slow under concurrent load, because no other request could be serviced while podman was running.

Wrap every `run_podman()` call in `await asyncio.to_thread(...)`, matching the pattern already used by `get_container_public_ip`. This offloads the blocking subprocess to the default thread-pool executor so the event loop stays responsive.

To verify, hit the API to create a new browser and while doing that, load the web UI (which will invoke the API to enumerate the browsers).

Before this change, the web UI feels very slow since its API call is blocked.

After this change, everything run asynchronously and there is no more blocking of or due to the web UI.

### DIFF
--- a/getgather/browsers/podman_browsers.py
+++ b/getgather/browsers/podman_browsers.py
@@ -27,7 +27,7 @@ def run_podman(args: list[str]) -> subprocess.CompletedProcess[str]:
 
 async def get_host_port(container_name: str, container_port: int) -> int | None:
     try:
-        result = run_podman(["port", container_name, str(container_port)])
+        result = await asyncio.to_thread(run_podman, ["port", container_name, str(container_port)])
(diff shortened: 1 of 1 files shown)


bin-ario (approved):
Would asyncio.create_subprocess_exec be better ?

ariya:
> Would asyncio.create_subprocess_exec be better ?

I believe. I will do a followup refactoring using that!

PR #1352, opened 2026-06-22
feat: add CDP screencast recording to MP4 for session investigation

oshimayoan:
Records live Chrome Fleet browser sessions as MP4 for post-incident investigation. When an MCP consumer passes `x-record: 1`, the server captures CDP screencast frames and encodes them to H.264 MP4 via ffmpeg on session end. Recordings are stored to local disk in dev or Fly Tigris in production.


faisalive:
@oshimayoan can you provide some evidence

oshimayoan:
@faisalive I can, but since this repo is public, I'm not sure about uploading the recording that contains our credentials

oshimayoan:
@faisalive You can find the recording here: https://heyario.slack.com/archives/C049DGC1D2M/p1782130329124889?thread_ts=1781656872.228029&cid=C049DGC1D2M

getgather/recording.py
```diff
@@ -0,0 +1,247 @@
...
+class _ActiveRecording:
+    meta: RecordingMeta
+    page: zd.Tab
+    frames_dir: Path
+    frame_count: int
+    timeout_task: asyncio.Task[None]
+    started_ts: float
+
+
+_active: dict[str, _ActiveRecording] = {}
```
bin-ario, line 47: does this work if the service runs multiple instances?

It would make sense to implement the recording inside chrome-live image, similar to residential proxy, browser-trace, so that the remotebrowser or fleet just needs to pass a flag to activate it

ariya: Agree with @bin-ario. In the spirit of ratchet model, it's easier to add this feature to browser-trace first (https://github.com/remotebrowser/browser-trace). In addition, test it directly first with chrome-live/chromium-live container, without involving Remote Browser.

oshimayoan: Ah, I see. I'll try to explore it on those codebases first, that is indeed the better place to place the recording implementation

faisalburhanudin:
Closed, moved to https://github.com/remotebrowser/browser-trace/pull/2

PR #1353, opened 2026-06-22
feat(daytona): warm pool for incognito browsers

faisalive:
# feat(daytona): warm pool for incognito browsers

## Problem

Every incognito browser request cold-creates a Daytona sandbox from snapshot. Live-fleet logs showed ~23s from request to usable browser, dominated by a ~9s sandbox create + boot.

## Change

Add a **warm pool** of pre-booted ephemeral sandboxes in `DaytonaBackend`. Incognito requests claim a ready spare instead of cold-creating one. State lives entirely in Daytona labels — the backend keeps no in-memory map, so it survives restarts.

- **`create_browser` keeps its signature.** For an incognito (`E`-prefixed) id with a non-empty pool, it claims a spare and records the binding as a `claimed_as=<requested-id>` label on the spare sandbox. Every later lookup (`get_browser`, `delete_browser`, `browser_exists`, `get_cdp_base_url`, `get_live_view_url`) resolves through that label. Per-user ids and an empty pool cold-create unchanged.
- **Stateless.** The spare keeps its own immutable `chromium-spare-<rand>` name; the claim replaces its labels (`set_labels` is a full replace) — atomically dropping `pool=spare` and recording the binding. No `_id_map`, no `_spares` set. Resolution survives a process restart and reads consistently across instances.
- **Proxy applied at claim**, not baked into spares — each incognito keeps a fresh residential IP.
- **Auto-refill.** Each claim kicks a background `_reconcile_pool` (own lock, so slow spawns never block a claim) to backfill to `POOL_SIZE`. The 5-min `cleanup_idle` sweep also reconciles, so the pool self-heals (e.g. spares Daytona auto-stopped after idle).
- **`POOL_SIZE` via `DAYTONA_INCOGNITO_POOL_SIZE`** (default 2). `0` disables → exact prior cold-create behavior.
- `dpage` is unchanged from main; the pool is fully contained in the backend.

## Result (measured against live fleet)

Request → usable browser: **~23s → ~13s**. The 9s cold-boot is replaced by a ~1s claim.

```
18:57:29  /cdp/Evy9dgnq  "Browser not found — launching"
18:57:30  Claimed pooled spare spare-… for incognito browser Evy9dgnq   ← ~1s
18:57:42  Start with an ephemeral browser Evy9dgnq
```

## Cost

`POOL_SIZE` spares run 24/7 = net-new compute (~$36–73/mo each; 2 ≈ $72–146/mo). Claimed sandboxes keep the existing 15m auto-stop / 60m auto-delete, so no extra cost vs today. `POOL_SIZE=0` is the kill switch.

## Tradeoffs / follow-ups

- **Per-lookup cost:** an incognito `get/delete/exists/cdp` now does one Daytona label query (~100–300ms). Per-user ids and disabled-pool skip it entirely.
- **Cross-instance claim races** are possible (last-writer-wins on the label) and self-heal; the in-process `_pool_lock` covers the single-process deployment. A future deployment needing hard multi-instance safety would add a CAS store.
- **Burst > `POOL_SIZE`** falls back to cold-create until background refill catches up.
- **`"E"` prefix is shared by convention** between dpage (mints it) and the backend (claims on it); a comment flags the contract.
- **Known, not fixed here:** `get_browser` re-applies the proxy on query, duplicating the claim-time config (~4s + a misleading "IP unchanged" warning). Pre-existing; the pool just made it visible. Worth a follow-up.

## Files

- `daytona_browsers.py` — pool: `_claim_spare_for`, `_resolve`, `_reconcile_pool`, `_spawn_spare`, `_iter_sandboxes`, label-based binding.
- `backend.py` — `startup()` added to the protocol.
- `fleet_browsers.py`, `podman_browsers.py` — no-op `startup()`.
- `settings.py` — `DAYTONA_INCOGNITO_POOL_SIZE`.
- `main.py` — `await backend.startup()` in lifespan.

## Test

- ruff format/lint, pyright (clean on changed files), 64 unit tests pass.
- Verified against live Chrome Fleet: pool claim + label resolution + auto-refill confirmed in logs.

### DIFF
--- a/getgather/browser.py
+++ b/getgather/browser.py
@@ -222,30 +222,6 @@ async def create_remote_browser(
     return browser
 
 
-async def acquire_incognito_remote_browser(
-    target_domain: str | None = None,
-) -> tuple[str, zd.Browser]:
-    """Acquire an ephemeral browser from the fleet (warm pool when available), returning its
-    server-minted browser_id and a connected zendriver Browser.
-
-    Unlike create_remote_browser, the id is assigned by the fleet so a pooled sandbox can be
-    handed out as-is instead of cold-creating one per request.
-    """
-    headers = build_chromefleet_headers(target_domain=target_domain)
-    response = await call_chromefleet_api(
-        "POST", path="/api/v1/browsers-incognito", headers=headers
-    )
-    if response is None:
-        raise RuntimeError("Failed to acquire incognito browser from Chrome Fleet")
-    browser_id = str(response.json()["browser_id"])
-    logger.info(f"Acquired incognito browser: {browser_id}")
-    cdp_websocket_url = _setup_cdp_url(browser_id)
-    browser = await _create_browser_from_cdp_websocket(
-        browser_id=browser_id, websocket_url=cdp_websocket_url, target_domain=target_domain
-    )
-    return browser_id, browser
-
-
 async def terminate_remote_browser(browser: zd.Browser) -> None:
     """Terminate an existing remote Chrome via ChromeFleet."""
     browser_id = cast(str, browser.id)  # type: ignore[attr-defined]
--- a/getgather/browser.py
+++ b/getgather/browser.py
@@ -222,6 +222,30 @@ async def create_remote_browser(
     return browser
 
 
+async def acquire_incognito_remote_browser(
+    target_domain: str | None = None,
+) -> tuple[str, zd.Browser]:
+    """Acquire an ephemeral browser from the fleet (warm pool when available), returning its
+    server-minted browser_id and a connected zendriver Browser.
+
+    Unlike create_remote_browser, the id is assigned by the fleet so a pooled sandbox can be
+    handed out as-is instead of cold-creating one per request.
+    """
+    headers = build_chromefleet_headers(target_domain=target_domain)
+    response = await call_chromefleet_api(
+        "POST", path="/api/v1/browsers-incognito", headers=headers
+    )
+    if response is None:
+        raise RuntimeError("Failed to acquire incognito browser from Chrome Fleet")
+    browser_id = str(response.json()["browser_id"])
+    logger.info(f"Acquired incognito browser: {browser_id}")
+    cdp_websocket_url = _setup_cdp_url(browser_id)
+    browser = await _create_browser_from_cdp_websocket(
+        browser_id=browser_id, websocket_url=cdp_websocket_url, target_domain=target_domain
+    )
+    return browser_id, browser
+
+
 async def terminate_remote_browser(browser: zd.Browser) -> None:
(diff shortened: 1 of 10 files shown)


ariya:
Ideally, we want to avoid warm pool since it adds some complexity (in debugging), and also it defeats the point of using something like Daytona. Thus, before we embrace on pooling support, @faisalive  do you have a detailed breakdown of that 23 seconds vs 9 seconds processing delay?

faisalive:
> Ideally, we want to avoid warm pool since it adds some complexity (in debugging), and also it defeats the point of using something like Daytona. Thus, before we embrace on pooling support, @faisalive do you have a detailed breakdown of that 23 seconds vs 9 seconds processing delay?

Yep, I think my previous timing test was not accurate. After running several tests, I noticed that sometimes creation is slow and sometimes it's fast.

Here are my latest test results comparing the deployed remotebrowser (FlyFleet) and remotebrowser-daytona:

remotebrowser:
24715ms | 21937ms | 15350ms | 15865ms | 20338ms

remotebrowser-daytona:
26718ms | 20542ms | 25531ms | 20009ms | 17543ms

Based on these results, there doesn't seem to be a significant difference between the two.

On the other hand, if we look at our Test Manager execution times, Daytona appears to be faster than FlyFleet. In fact, Daytona performs about the same as Lambda.

| Workload     | Metric           | Demo    | Daytona | Lambda |
|--------------|------------------|----------|----------|----------|
| Headline Hub | Average Duration | 70.0s   | **43.3s** | 43.8s |
|              | Fastest Run      | 57s     | 37s     | **33s** |
|              | Slowest Run      | 82s     | 52s     | 62s |
|              | Consistency      | Low     | **High** | Medium |
| Page Turner  | Average Duration | 101.7s  | 76.9s   | **76.8s** |
|              | Fastest Run      | 77s     | 71s     | **57s** |
|              | Slowest Run      | 152s    | **92s** | 132s |
|              | Consistency      | Low     | **High** | Medium |
| Overall      | Average Duration | 85.9s   | **60.1s** | 60.3s |

bin-ario:
Thanks, @faisalive . Your latest experiment aligns with my experience.
One thing I think is worth implementing is the best-of-N strategy, i.e., launching N=3 daytona sandboxes and picking the fastest one with IP setup (killing the other N-1), that should help deal with the daytona latency variance

faisalive:
> Thanks, @faisalive . Your latest experiment aligns with my experience. One thing I think is worth implementing is the best-of-N strategy, i.e., launching N=3 daytona sandboxes and picking the fastest one with IP setup (killing the other N-1), that should help deal with the daytona latency variance

I see, I'll try that. Thankyou

PR #1355, opened 2026-06-23
Make the entire podman process natively async

ariya:
This is a follow-up to PR #1350

Replace `asyncio.to_thread(subprocess.run)` with the true `asyncio.create_subprocess_exec` to avoid blocking the event loop thread pool and simplify call sites.


ariya:
This does not seem to work, closing it for now.

PR #1356, opened 2026-06-23
Make the entire podman process natively async

ariya:
This is a follow-up to PR #1350.

Replace `asyncio.to_thread(subprocess.run)` with the true `asyncio.create_subprocess_exec` to avoid blocking the event loop thread pool and simplify call sites.

It's the second attempt, after the earlier PR #1355 failed.

To verify, run Remote Browser first. Then launch Headline Hub pointing to this Remote Browser instance. All the functionalities from Headline Hub should work as expected.

### DIFF
--- a/getgather/browsers/podman_browsers.py
+++ b/getgather/browsers/podman_browsers.py
@@ -15,19 +15,49 @@ from getgather.config import settings
 DOCKER_INTERNAL_HOST = "172.17.0.1"
 
 
-def run_podman(args: list[str]) -> subprocess.CompletedProcess[str]:
+# uvloop's child process watcher can cause asyncio.create_subprocess_exec to hang
+# indefinitely. When uvloop is the active event loop, fall back to running the
+# subprocess in a thread via asyncio.to_thread to avoid the deadlock.
+def _is_uvloop() -> bool:
+    return type(asyncio.get_event_loop()).__module__.startswith("uvloop")
+
+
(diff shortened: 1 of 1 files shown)


getgather/browsers/podman_browsers.py
```diff
@@ -15,19 +15,49 @@
...
+
+async def run_podman(args: list[str]) -> subprocess.CompletedProcess[str]:
     cmd = ["podman"]
     if settings.CONTAINER_HOST:
         cmd.append("--remote")
     cmd.extend(args)
-    return subprocess.run(
-        cmd, capture_output=True, text=True, check=True, encoding="utf-8", errors="replace"
+
+    if _is_uvloop():
```
ariya, line 31: This is the key to the solution, which was missing in my earlier PR #1355.

bin-ario: interesting, we can try this

the same asyncio.create_subprocess_exec worked fine in the old gateway codebase https://github.com/remotebrowser/mcp-gateway/blob/main/src/container/engine.py#L387

bin-ario (approved):

getgather/browsers/podman_browsers.py
ariya: Not sure how, that `run_cli` seems to be about launching a process one-off, whereas `uvicorn.run()` (which runs in the uvicorn loop) is for a long-running process (i.e. a server). But let me keep digging.

PR #1359, opened 2026-06-24
feat: Best-of-N Daytona Sandbox Creation

faisalive:
On cold-create, launch N sandbox candidates in parallel and take the first to fully pass (started + verified proxy IP change). Losers are deleted in the background.

`DAYTONA_BEST_OF_N=3` (default). Set to 1 to disable.

## Why

Single cold-create is slow and can land a bad proxy. Racing N candidates cuts p99 latency and avoids proxy failures blocking the user.

## What changed

**Identity moves from names to labels.** Sandboxes now carry a `browser_id=<id>` label. Names are random (`prefix-<id>-<uuid8>`). All lookups use a label query. This allows N same-id sandboxes to coexist during the race.

**`_best_of_n`** — runs N candidates concurrently, takes the first full-pass winner, cancels the rest. If all fail proxy verification, falls back to fastest-started with a warning. Schedules fire-and-forget teardown of losers + orphan sweep via label query.

**`create_browser`** — pre-checks for an existing started sandbox (fast-path); otherwise best-of-N or single path depending on `N`.

**Migration:** existing sandboxes (no `browser_id` label) become invisible at deploy and auto-deleted by Daytona within 75 min. Re-auth is the recovery path.

## Tests

6 unit tests covering: winner selection, proxy-fail fallback, zero-create/zero-start raises, N=1 single path, existing-sandbox fast-path.


faisalive:
# Best-of-N vs Base — Timing Benchmark

## Setup

- **Base**: single `create()` → proxy config → IP verification
- **BoN (N=3)**: 3 candidates race in parallel; first to fully pass wins
- 5 concurrent users firing simultaneously in both batches
- Both batches run at the same time (competing for Daytona resources)

---

## Without Proxy — Pure `create()` speed

| User | Base | BoN (N=3) |
|------|------|-----------|
| 1 | 1.6s | 1.6s |
| 2 | **22.6s** 🔥 | 1.6s |
| 3 | 1.5s | 1.6s |
| 4 | 1.7s | 1.6s |
| 5 | 1.7s | 1.6s |
| **Avg** | **5.8s** | **1.6s** |
| **Wall (all done)** | **22.6s** | **1.6s** |

Base user 2 hit a stuck machine — blocked the entire batch for 22.6s.
BoN escaped it. **Wall clock 14× faster.**

---

## With Proxy — `create()` + proxy config + IP verification

| User | Base | BoN (N=3) |
|------|------|-----------|
| 1 | 10.4s | 8.2s |
| 2 | 9.2s | 8.3s |
| 3 | 9.1s | 8.2s |
| 4 | 9.6s | 6.8s |
| 5 | **37.8s** 🔥 | 8.1s |
| **Avg** | **14.7s** | **7.9s** |
| **Wall (all done)** | **37.9s** | **8.3s** |

Base user 5 stuck through full proxy chain — 37.8s.
BoN escaped. **Wall clock 4.5× faster.**

---

## Key Finding

`create()` latency has high variance: **1.5s (fast) → 22–38s (stuck machine)**.
Single path is at the mercy of whichever machine Daytona assigns.
BoN races N=3 and takes the fastest:

```
worst case = min(slow, slow, fast) ≈ fast
```

Stuck-machine scenario is not rare — appeared in **2 out of 10 sequential runs**
and in **both parallel batch runs**.

ariya:
@faisalive I think it will be useful to hold off this PR for now, until every client is using the REST API to request a remote browser with automatic naming. That way, there's no need to store an internal mapping. Best-of-N logic is basically simplified to:
1. create N remote browsers (all with automatic/random name)
2. check each one, determine the best
3. return that one to the caller

PR #1365, opened 2026-06-25
feat(target): implement get_purchases with API interception and curation

anazaim77:
## Summary

- **`target.py`**: New `get_purchases` tool fetches Target online purchase history — intercepts first page via JS `window.fetch` monkey-patch to capture both data and `x-api-key` from live request headers, then fetches remaining pages and parallel-fetches all order details via `Promise.all` in-browser
- **`target.py`**: Returns raw API response objects — no client-side normalization, full order detail JSON preserved
- **`target-purchase.html`**: Trimmed pattern — removed duplicate signed-in selector and unused JSON converter
- **`target-signin-otp.html`**: New OTP pattern for Target sign-in flow
- **`target-signin-error.html`**: Updated error pattern with `gg-error` codes
- **`target-maybe-later.html`**: New pattern for Target "maybe later" prompt
- **`mcp-tools.yaml`**: Removed `get_purchases_in_store` (unimplemented)

## Test Plan

- [ ] Unit tests pass: `uv run pytest -m "not api and not webui and not mcp and not distill"`
- [ ] MCP integration test — requires live server + Chrome Fleet + `TARGET_EMAIL` / `TARGET_PASSWORD` env vars:
  - [ ] Unauthenticated call returns `{url, signin_id}` sign-in prompt
  - [ ] After sign-in completes, `check_signin` returns `{status: "SUCCESS", completed: true}`
  - [ ] Re-call `target_get_purchases` returns `{target_purchases: [...]}` list
  - [ ] Each purchase contains full raw order detail object
  - [ ] Multi-page accounts: orders beyond page 1 (>10 orders) are included

## Result

image


getgather/mcp/target.py
```diff
@@ -1,26 +1,201 @@
...
+from getgather.browser import zen_navigate_with_retry
 from getgather.mcp.dpage import remote_zen_dpage_with_action
 from getgather.mcp.registry import MCPTool
+from getgather.zen_actions import parse_response_json
 
 target_mcp = MCPTool.registry["target"]
 
+BASE_URL = "https://www.target.com"
+API_BASE = "https://api.target.com"
+X_API_KEY = "ff457966e64d5e877fdbad070f276d18ecec4a01"
```
faisalive: this one hardcoded?

getgather/mcp/target.py
```diff
@@ -1,26 +1,201 @@
...
+        if not order_numbers:
+            return {"target_purchases": []}
+
+        try:
+            details = await asyncio.wait_for(_fetch_all_details(page, order_numbers), timeout=60.0)
+        except asyncio.TimeoutError:
+            logger.warning("Target: detail fetch timed out")
+            details = []
+
+        return {"target_purchases": [_curate_order(d) for d in details]}
```
faisalive, line 139: Why not just return the raw `details` and let client do the normalization? That way we can have a log of full  raw details inside `MCP Response` and client have flexibility with the data

getgather/mcp/target.py
```diff
@@ -1,26 +1,201 @@
...
-            "message": "Successfully signed in and reached order history page. Purchase history extraction not yet implemented.",
-        }
+    async def action(page: zd.Tab, browser: zd.Browser) -> dict[str, Any]:
+        logger.info("Target: signed in, fetching online purchase history")
+
+        order_numbers: list[str] = []
+
+        async with page.expect_response(".*/guest_order_aggregations/v1/order_history.*") as resp:
+            await zen_navigate_with_retry(page, f"{BASE_URL}/orders", wait_for_ready=False)
+            page1: dict[str, Any] = await parse_response_json(resp, {}, "target order list page 1")
```
faisalive: Maybe we can use eval instead of `page.expect_response`

getgather/mcp/target.py
anazaim77: updated

getgather/mcp/target.py
anazaim77: sure, updated

getgather/mcp/target.py
anazaim77: done this

faisalive (approved):

PR #1372, opened 2026-07-01
Allow creating a browser with an auto-generated id

ariya:
When the browser id is omitted, the server generates one automatically. This is convenient for certain clients that consume Remote Browser, since they no longer need to pick a unique id themselves.

Example:

```bash
$ curl -X POST http://localhost:23456/api/v1/browsers
{"browser_id":"Bi8rzsq9u","container_name":"chromium-Bi8rzsq9u",
"status":"created","ip":"73.158.108.127"}
```

### DIFF
--- a/getgather/browsers/router.py
+++ b/getgather/browsers/router.py
@@ -9,9 +9,11 @@ from fastapi import APIRouter, HTTPException, Request, WebSocket, WebSocketDisco
 from fastapi.responses import HTMLResponse, JSONResponse
 from fastapi.websockets import WebSocketState
 from loguru import logger
+from nanoid import generate
 from websockets.exceptions import ConnectionClosed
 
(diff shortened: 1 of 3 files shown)


getgather/browsers/router.py
```diff
@@ -207,6 +209,22 @@ async def remote_to_client() -> None:
             await client_ws.close(code=4500, reason="Internal proxy error")
 
 
+@router.post("/api/v1/browsers")
+async def create_browser_auto(request: Request) -> dict[str, Any]:
+    browser_id = "B" + generate(FRIENDLY_CHARS, 8)
```
ariya, line 214: The `B` prefix is intentional, it stands for (surprise) **B** rowser.

bin-ario (approved):

PR #1375, opened 2026-07-06
feat(captcha): add logging for triggered captcha type

faisalburhanudin:
## Why

Determining which captcha hit the most lets us focus solver effort on the captchas that actually block us the most.

## How
When a captcha stop pattern fires, sniff the live DOM for each vendor's fingerprint (script/iframe URLs and known element ids) and emit the type to Logfire (`captcha detected` span with `captcha_type`/`hostname` / `pattern`).

### Counting in Logfire

```sql
select attributes->>'captcha_type' as type, count(*)
from records
where message = 'captcha detected'
group by type order by count(*) desc
```

## Tests

script to test

```
import asyncio, os
os.environ.setdefault("CHROMEFLEET_URL", "http://localhost:1")  # unused; satisfies config import
import zendriver as zd
from getgather.zen_distill import classify_captcha

BROWSER = "/Applications/Brave Browser.app/Contents/MacOS/Brave Browser" 
DEMOS = {
    "recaptcha_v2": "https://www.google.com/recaptcha/api2/demo",
    "hcaptcha": "https://accounts.hcaptcha.com/demo",
    "cloudflare": "https://demo.turnstile.workers.dev/",
    "geetest": "https://www.geetest.com/en/demo",
    "aws_waf": "https://dcz1t9chpw9yt.cloudfront.net/image-captcha",
    "image_captcha": "https://dcz1t9chpw9yt.cloudfront.net/image-captcha"
}

async def main():
    for expected, url in DEMOS.items():
        browser = await zd.start(browser_executable_path=BROWSER, headless=True)
        try:
            page = await browser.get(url)
            await asyncio.sleep(7)  # let the widget inject its iframe
            got = await classify_captcha(page)
        finally:
            await browser.stop()
        print(f"[{'PASS' if got == expected else 'FAIL'}] {expected} -> {got}  {url}")

asyncio.run(main())
```

Ran `classify_captcha` against real challenge pages in a local Chromium (zendriver), one fresh browser per URL. 

| Expected       | Result | Page                                               |
| -------------- | ------ | -------------------------------------------------- |
| `recaptcha_v2` | PASS   | https://www.google.com/recaptcha/api2/demo         |
| `hcaptcha`     | PASS   | https://accounts.hcaptcha.com/demo                 |
| `cloudflare`   | PASS   | https://demo.turnstile.workers.dev/                |
| `geetest`      | PASS   | https://www.geetest.com/en/demo                    |
| `aws_waf`      | PASS   | https://dcz1t9chpw9yt.cloudfront.net/protected     |
| `image_captcha`    | PASS   | https://dcz1t9chpw9yt.cloudfront.net/image-captcha |

### DIFF
--- a/getgather/zen_distill.py
+++ b/getgather/zen_distill.py
@@ -192,36 +192,7 @@ async def get_error(distilled: str) -> str | None:
     return None
 
 
-CAPTCHA_CLASSIFIER_JS = r"""
-(() => {
-  try {
-    const has = (sel) => { try { return !!document.querySelector(sel); } catch (e) { return false; } };
-    const srcs = [];
-    document.querySelectorAll('script[src], iframe[src], link[href]').forEach((el) => {
-      const v = el.getAttribute('src') || el.getAttribute('href') || '';
-      if (v) srcs.push(v);
-    });
-    let html = '';
-    try { html = document.documentElement.outerHTML; } catch (e) {}
-    const hay = (srcs.join(' ') + ' ' + html).toLowerCase();
-
-    if (hay.includes('awswaf.com') || hay.includes('awswafintegration') || hay.includes('awswafcaptcha')) return 'aws_waf';
-    if (hay.includes('arkoselabs.com') || hay.includes('funcaptcha.com') || has("iframe[title='verification puzzle']") || has('#cvf-aamation-challenge-iframe')) return 'arkose_funcaptcha';
-    if (hay.includes('hcaptcha.com') || has('.h-captcha') || has('iframe[src*="hcaptcha"]')) return 'hcaptcha';
-    if (hay.includes('recaptcha/enterprise')) return 'recaptcha_enterprise';
-    if (/recaptcha\/api\.js\?[^\s"']*render=/.test(hay)) return 'recaptcha_v3';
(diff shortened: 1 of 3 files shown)


getgather/zen_distill.py
```diff
@@ -583,6 +650,9 @@ async def run_distillation_loop(
                 current = match
 
                 if await terminate(distilled):
+                    error_code = await get_error(distilled)
+                    if error_code == "captcha":
+                        await report_captcha(page, pattern_name=match.name, hostname=hostname)
```
faisalive, line 626: Not all `gg-error="captcha"` are actually a real captcha, is that ok?

Image

faisalburhanudin: It should be fine. Non-captcha `gg-error="captcha"` pages fall into `unknown`, which is intended as a best-effort.

getgather/zen_distill.py
```diff
@@ -191,6 +192,72 @@ async def get_error(distilled: str) -> str | None:
     return None
 
 
+CAPTCHA_CLASSIFIER_JS = r"""
```
ariya: While this is fine for now, I think eventually it's better to source this from an external file (so we can properly check for syntax error, formatting guide, etc).

PR #1378, opened 2026-07-07
Remove MCP OAuth authentication

faisalive:
## Summary
- Drop OAuth-based MCP auth entirely: delete `getgather/auth/provider.py` (CustomOAuthProvider/CustomTokenVerifier) and `getgather/auth/settings.py` (auth_enabled gating), gut `getgather/auth/auth.py` to a stub `get_auth_user()` with no per-request identity, remove `setup_mcp_auth()` call from `main.py`.
- Drop now-unused `py-key-value-aio[redis]` dependency and stale `OAUTH_*`/`FIRST_PARTY_*` env vars from `.env.template`.
- Fix a regression introduced by the removal: `browser_id` in `getgather/mcp/dpage.py` was still keyed off `get_auth_user().user_id`, which is now a process-wide constant — different callers would collapse onto the same persistent authenticated browser. Now keys off `mcp-session-id` first (falling back to the old hostname-based id when absent), so distinct client sessions get distinct browsers again.

## Test plan
- [x] `uv run pyright .` — 0 errors
- [x] `uv run ruff check . && uv run ruff format --check .` — clean
- [x] `uv run pytest -m "not api and not webui and not mcp and not distill"` — 65 passed

Screenshot 2026-07-08 at 13 21 30


bin-ario:
does it necessary to keep the `AuthUser` class and `get_auth_user` function?
prefer removing them if we don't do auth anymore

faisalive:
> does it necessary to keep the `AuthUser` class and `get_auth_user` function? prefer removing them if we don't do auth anymore

yap, removed it also

broerjuang (approved):
LGTM

PR #1396, opened 2026-07-13
Gradually transition from npm scripts to a Makefile

ariya:
Since this is purely a Python project, package.json and its npm scripts are not strictly necessary. Use Makefile targets instead for the various operations (running the dev server, formatting code, etc).

During this transition, npm scripts are still supported, though internally they map to the corresponding Makefile targets. For example, `npm run dev` is effectively just `make dev`.

Note that Node.js—specifically its `npx` utility—is still required to run Prettier and to power commit hooks via Husky (these will be migrated later).

To verify, run the various npm scripts and they should still work. All CI workflows should continue to pass as usual.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -33,7 +33,7 @@ jobs:
         uses: ./.github/actions/prepare-frontend
 
       - name: Check frontend formatting
-        run: npm run frontend:check:format
+        run: make check-frontend-format
 
   backend-types:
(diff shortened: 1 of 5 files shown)


bin-ario (approved):
justfile could be a more modern alternative to makefile

ariya:
> justfile could be a more modern alternative to makefile

Definitely, though my goal was to avoid an extra dependency (`make` is most likely already preinstalled everywhere).

PR #1400, opened 2026-07-14
feat: Best-of-N Daytona sandbox creation

faisalive:
On cold-create via `POST /api/v1/browsers`, launch N sandbox candidates in parallel and take the first that fully succeeds (started + proxy verified). Losers are deleted in the background.

`DAYTONA_BEST_OF_N=3` (default). Set to `1` to disable the race.

## Why

Single cold-create is slow and can land a bad proxy. Racing N candidates cuts the p99/tail latency and avoids a bad proxy blocking the user.

## Proxy is mandatory

If a proxy is configured for the browser, it MUST apply and change the egress IP. Otherwise the browser is rejected — we never hand back an unproxied browser:

- **best-of-N**: a candidate only counts as a winner after it starts AND passes proxy verification. If **every** candidate fails verification, best-of-N raises `ProxyVerificationError` (→ 500) instead of falling back to an unverified browser, so the client can retry.
- **single path** (`create_browser` / `get_browser`): the error propagates the same way.
- If **no** proxy is configured for the browser, verification is skipped (proxy is not required).

## Design — no identity change

- Each candidate is a normal browser with its own **server-assigned `browser_id`** and the existing **`chromium-<id>`** sandbox name. Ids are unique, so names never collide during the race.
- No `browser_id` label, no local mapping, no rename. `get`/`delete`/`cdp`/`live_view` lookups are untouched.
- The winner's id is returned to the caller and is authoritative.

Scope: only `POST /api/v1/browsers` (server assigns the id). `POST /api/v1/browsers/{browser_id}` is unchanged — the caller already committed to an id, so there is no winner to swap in; that path does **not** race.

## What changed

- **`backend`** — `new_browser_id()` helper + `create_browser_auto()` on the `Backend` protocol.
- **`router`** — `POST /api/v1/browsers` delegates id assignment to the backend.
- **`daytona`** — `_configure_remote_sandbox` raises `ProxyVerificationError` on a failed mandatory proxy (no-op when none configured). `_best_of_n` races candidates via `as_completed`, takes the first to complete, raises if all fail, and fires `_cleanup_losers`.
- **`_cleanup_losers`** — retries the delete. A cancelled candidate is usually still *starting*, so Daytona rejects the delete with `"Sandbox state change in progress"`; retry until it settles (Daytona's `auto_delete_interval` is the final backstop).
- **`fleet`** — forwards to the upstream auto endpoint (upstream owns the race). **`podman`** — single-create, no race.

## Tests

Unit: first-to-complete winner selection, no-proxy-verified raises, all-fail raises, N=1 single path, cleanup deletes losers only. `make check` + `make test` green.

## Live verification (real Daytona, proxy enabled)

5 requests each to `POST /api/v1/browsers` with `x-origin-ip`:

| stat | N=1 (single) | N=3 (best-of) | Δ |
|------|-------------|---------------|---|
| min | 8.03s | 7.61s | −5% |
| mean | 11.93s | 8.26s | **−31%** |
| median | 9.95s | 8.16s | −18% |
| **max (tail)** | **22.23s** | **9.42s** | **−58%** |

N=1 hit a 22s bad-proxy request; best-of-3 capped max at ~9.4s. The win is killing the tail. Logs confirm each request ran a `Best-of-3 sandbox race` and the winner was frequently **not** the first-listed candidate → selection actually chooses.

**Loser cleanup:** the retry fix was found and validated here — an early pass leaked 3/10 losers (`state change in progress`); after the fix, **10/10 (and 6/6 on re-verify) losers deleted, 0 gave up**. All test sandboxes cleaned up afterward.

### DIFF
--- a/getgather/browsers/daytona_browsers.py
+++ b/getgather/browsers/daytona_browsers.py
@@ -13,14 +13,9 @@ from daytona import (
 )
 from loguru import logger
 
-from getgather.browsers.backend import BROWSER_NAME_PREFIX, BrowserNotFound
+from getgather.browsers.backend import BROWSER_NAME_PREFIX, BrowserNotFound, new_browser_id
 from getgather.browsers.residential_proxy import get_proxy_config
 from getgather.config import settings
-
-
-class ProxyVerificationError(Exception):
-    """Raised when egress IP is unchanged after proxy configuration."""
-
 
 CDP_PORT = 9222
 VNC_PORT = 8080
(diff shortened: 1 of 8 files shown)


getgather/browsers/daytona_browsers.py
```diff
@@ -245,6 +249,82 @@ async def get_live_view_url(self, browser_id: str) -> str | None:
...
+            try:
+                browser_id, info, proxy_ok = await coro
+            except Exception as e:
+                logger.warning(f"Best-of-N candidate failed: {type(e).__name__}: {e}")
+                continue
+            if proxy_ok:
+                winner = (browser_id, info)
+                break
+            if fallback is None:
+                fallback = (browser_id, info)
```
bin-ario: what's the use of fallback? if proxy doesn't work, would it be better to fail, then the client can retry if necessary?

faisalive: You're right, I've removed it, thankyou!

bin-ario (approved):

PR #1401, opened 2026-07-15
fix: assign browser id client-side in FleetBackend.create_browser_auto

faisalive:
## Problem

Sentry [MCP-GETGATHER-82X](https://heyario.sentry.io/issues/7611693622/) — `HTTPException: Unable to start browser!` (16 occurrences, 3 users, escalating on `demo`).

Root cause is a nested `HTTPStatusError: 405 Method Not Allowed` for `POST /api/v1/browsers`. `FleetBackend.create_browser_auto` POSTed to the fleet **collection** endpoint expecting server-assigned id, but the upstream Chrome Fleet only implements `POST /api/v1/browsers/{browser_id}`.

## Fix

Assign the id client-side via `new_browser_id()` and forward to the per-id endpoint — matching `PodmanBackend` and `DaytonaBackend`, which already do this.

`make check` passes.

Fixes MCP-GETGATHER-82X

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -17,6 +17,9 @@ MAX_IDLE_MINUTES=15
 DAYTONA_API_KEY=
 DAYTONA_API_URL=
 DAYTONA_SNAPSHOT=
+# Best-of-N cold-create: race this many sandbox candidates and keep the first proxy-verified one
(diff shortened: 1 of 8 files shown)


faisalive:
Superseded by clean fix branched off latest main (Best-of-N merged as #1400).

PR #1404, opened 2026-07-15
fix: correct Daytona proxy verification and stop reconfiguring proxy on GET

faisalive:
Two related fixes to Daytona proxy handling (follow-up to #1400).

## 1. Proxy verification folded distinct outcomes into one misleading raise

`_configure_remote_sandbox` verified the proxy with:

```python
if ip_before and ip_after and ip_before != ip_after:
    return
raise ProxyVerificationError(f"...IP unchanged after proxy: {ip_before}")
```

`_get_sandbox_public_ip` returns `None` on any curl/exec failure, so a `None` on **either** side short-circuits the `and` to False and raises a misleading **"IP unchanged"** — even when the proxy applied fine and the egress IP actually differed:

- `ip_before=None` — tinyproxy@8119 not yet serving on a freshly-created sandbox
- `ip_after=None` — the IP-check curl timing out (seen in prod as `UND_ERR_HEADERS_TIMEOUT` / "operation aborted")

A transient timeout thus failed **every** best-of-N candidate → `_best_of_n` all-fail → `ProxyVerificationError` to the client.

**Fix:**
- `ip_after is None` → raise a distinct **"IP check failed"** error (can't confirm either way), not "unchanged"
- "IP unchanged" only raises when **both** IPs are present and equal
- a missing `ip_before` no longer fails a working proxy

## 2. GET /api/v1/browsers/{id} no longer reconfigures the proxy

`get_browser` re-ran `_configure_remote_sandbox` on every GET carrying `x-origin-ip` — rewriting tinyproxy.conf, restarting tinyproxy (`s6-svc -r`), and re-running egress IP verification. On an existing browser that's pure waste: it restarts the proxy mid-session and (given fix #1) can 500 a plain read on a transient IP-check flake.

Proxy is now applied + verified **once at create**. GET is a cheap, idempotent read: look up the sandbox and return info.

> Note: browsers are single-purpose, so a persistent browser is not re-matched to a different proxy on later GETs. Re-create (POST) to change proxy.

## Tests
- `ip_before` missing → succeeds (regression)
- `ip_after` None → raises "IP check failed"
- IPs equal → raises "IP unchanged"
- `get_browser` never calls `_configure_remote_sandbox`, even with `origin_ip`

`make check` clean, all daytona tests pass.

### DIFF
--- a/getgather/browsers/daytona_browsers.py
+++ b/getgather/browsers/daytona_browsers.py
@@ -187,12 +187,12 @@ class DaytonaBackend:
     async def get_browser(
         self, browser_id: str, origin_ip: str | None, target_domain: str | None
     ) -> dict[str, Any]:
+        # Proxy is applied + verified once at create time; GET is a cheap read that never
+        # reconfigures (no tinyproxy restart, no egress re-check), so it stays idempotent and
+        # can't 500 on a transient IP-check flake.
         sandbox = await self._get(_sandbox_name(browser_id))
         if sandbox is None:
             raise BrowserNotFound(browser_id)
-        if origin_ip:
-            # Mandatory proxy: propagate ProxyVerificationError instead of returning unproxied.
-            await _configure_remote_sandbox(sandbox, browser_id, origin_ip, target_domain)
         return await self._get_info(sandbox)
 
     async def delete_browser(self, browser_id: str) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


oshimayoan (approved):
LGTM

faisalburhanudin (approved):

PR #1405, opened 2026-07-15
Generic best-of-N browser creation across all browser backends

ariya:
Unify browser creation policy across Daytona, Podman, and the external Chrome Fleet backend, and hoist it out of the backends into shared code.

- backend.py: add a shared `best_of_n` race helper, `_cleanup_losers`, `ProxyVerificationError`, and `_BestOfNBackend`/`_CleanupBackend` protocol slices so any backend (and minimal test fakes) can race N candidates and tear down the losers in the background.
- Backends now implement only per-browser CRUD keyed by a caller-supplied id; `create_browser_auto` is removed from the `Backend` protocol and each backend (Daytona/Podman/Fleet).
- Podman proxy is now mandatory-when-configured and strictly verified: `configure_container` returns apply success (bool) instead of raising inconsistently; `configure_remote_browser` raises ProxyVerificationError on apply failure, IP-check failure, or unchanged IP, mirroring Daytona so the best-of-N race can pick between podman candidates on proxy verification (not just container launch). No proxy -> no-op. `get_browser` is a cheap read: proxy is configured+verified once on create, never reconfigured on GET (even with x-origin-ip), so an IP-check flake can't 500 a GET.
- `.env.template`: rename `DAYTONA_BEST_OF_N` -> `BROWSER_BEST_OF_N`, documented as backend-agnostic.


getgather/browsers/router.py
```diff
@@ -208,12 +215,20 @@ async def remote_to_client() -> None:
...
+async def create_browser_auto_endpoint(request: Request) -> dict[str, Any]:
+    # Server-assigned id + best-of-N policy lives here (not per-backend): every backend only owns
+    # per-browser CRUD keyed by a caller-supplied id. When BROWSER_BEST_OF_N > 1 it races that many
+    # candidates via best_of_n and returns the winner; otherwise it creates a single browser directly.
     logger.info("Starting browser (server-assigned id)...")
     try:
         origin_ip = request.headers.get("x-origin-ip")
         target_domain = request.headers.get("x-target-domains")
-        browser_id, result = await backend.create_browser_auto(origin_ip, target_domain)
+        n = max(1, settings.BROWSER_BEST_OF_N)
```
bin-ario, line 226: instead of using the `BROWSER_BEST_OF_N` env variable directly, would be a good idea to add a `backend.best_of_n` method and use it here. this allows flyfleet to override the env variable and always return 1, because flyfleet already does its own best of N

bin-ario (approved):

ariya:
This is how it looks with Podman. Note the constructions of 3 browsers, one of them was declared as the winner (`Bywpb2zxf`) and the other ones were deleted. This winner was then used in the subsequent calls.

image

getgather/browsers/router.py
ariya: I like this idea of customizing the N depending on the backend!

Let me change the current default `BROWSER_BEST_OF_N` to 1, essentially making it no doing any N parallel browsers. Then I'll do a followup where each backend can specify the default N (e.g. Podman is best-of-5, Daytona is best-of-3, and external Fleet is best-of-1).

faisalive (approved):

PR #1407, opened 2026-07-16
feat: CloakBrowser support in Daytona backend (env-selected, single boot)

faisalive:
## What

Adds CloakBrowser support to the Daytona backend. Default stays **Chrome**; set `DAYTONA_ACTIVE_BROWSER=cloak` to boot CloakBrowser instead.

Depends on chrome-live PR https://github.com/remotebrowser/chrome-live/pull/22 (single image, both browsers, selected at boot by the `chromium` s6 service).

## How (env-selected, single boot)

- `settings.py`: `DAYTONA_ACTIVE_BROWSER` string setting (default `"chrome"`).
- `daytona_browsers._create`: pass `env_vars={"ACTIVE_BROWSER": <setting>}` at create.
- chrome-live `chromium/run` reads `$ACTIVE_BROWSER` on first boot (state file wins if present, for mid-session `switch-browser`; else env; else chrome).

A fresh create and a resumed start both boot the selected browser directly — no post-start `switch-browser` swap, no CDP re-wait; `_ensure` just starts a stopped sandbox. (An earlier revision swapped after start, which double-boots on cold create since Daytona `create()` auto-boots Chrome first — the env approach avoids it.)

## Validation (against the published snapshot)

**Latency** — create, N=5, best-of-N=1: cloak median **2.51s** vs chrome **2.57s** → no penalty (the old swap path cost ~+3.4s: chrome 2.87s → cloak 6.23s).

**Residential proxy on cloak** — egress IP changed `206.223.225.194` → `184.176.86.73` (HTTP 200; raises `ProxyVerificationError` if unchanged). Confirms cloak routes through tinyproxy same as chrome.

**Which browser booted** — noVNC live view is just pixels, can't distinguish; query CDP `/json/version` on the returned `cdp_url`:

```bash
curl -s "$CDP_URL/json/version" | jq '{Browser, "User-Agent", "V8-Version", "WebKit-Version"}'
```

| field | chrome | cloak |
|---|---|---|
| Browser | `Chrome/150.0.7871.124` | `Chrome/146.0.7680.177` |
| UA platform | `X11; Linux x86_64` | `Windows NT 10.0; Win64` |
| V8-Version | `15.0.245.19` | `14.6.202.31` |
| WebKit hash | `9261fd0a…` | `ae03f7fb…` |

Distinct V8 + WebKit build hashes = different binary. Cloak reports `Windows NT 10.0` on a Linux sandbox (`--fingerprint-platform=windows`) — the quick tell. It keeps `Chrome/<n>` on purpose (stealth), so compare platform/build, not `Browser` alone. Ground truth: `sudo cat /home/user/.active-browser` → `chrome` | `cloak`.

### DIFF
--- a/paseo.json
+++ b/paseo.json
@@ -1,0 +1,5 @@
+{
+  "worktree": {
+    "setup": "cp $PASEO_SOURCE_CHECKOUT_PATH/.env .env"
+  }
+}
--- a/paseo.json
+++ b/paseo.json
@@ -1,5 +1,0 @@
-{
-  "worktree": {
-    "setup": "cp $PASEO_SOURCE_CHECKOUT_PATH/.env .env"
-  }
-}
--- a/.env.template
+++ b/.env.template
@@ -22,6 +22,9 @@ MAX_IDLE_MINUTES=15
 DAYTONA_API_KEY=
 DAYTONA_API_URL=
 DAYTONA_SNAPSHOT=
+# Boot CloakBrowser (stealth build) instead of Chrome. Default off (Chrome). Requires a snapshot
+# whose image selects the browser via the ACTIVE_BROWSER env var.
+DAYTONA_CLOAK_BROWSER=
 
 # Residential proxy (Massive or Oxylabs) and MaxMind GeoIP. Consumed only by the podman backend,
 # which sets a per-browser upstream proxy and resolves geo from x-origin-ip. The Daytona backend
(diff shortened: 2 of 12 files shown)


broerjuang (approved):
LGTM

paseo.json
ariya, line 1: We probably don't want this.

faisalive: Yes 😅

PR #1414, opened 2026-07-23
Type checking with ty, in addition to pyright

ariya:
It's fast!

To verify, run `make typecheck`.

image

### DIFF
--- a/getgather/browsers/podman_browsers.py
+++ b/getgather/browsers/podman_browsers.py
@@ -31,6 +31,17 @@ def _is_uvloop() -> bool:
     return type(asyncio.get_event_loop()).__module__.startswith("uvloop")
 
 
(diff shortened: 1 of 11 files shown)


getgather/browsers/podman_browsers.py
```diff
@@ -31,22 +31,25 @@ def _is_uvloop() -> bool:
     return type(asyncio.get_event_loop()).__module__.startswith("uvloop")
 
 
+def _run_podman_sync(cmd: list[str]) -> subprocess.CompletedProcess[str]:
```
ariya, line 34: This is necessary to simplify the type propagation for the type checker. And it's a good refactoring anyway (for readability).

bin-ario (approved):

PR #1423, opened 2026-07-29
feat(dpage): add JSON sign-in step endpoint alongside the HTML dpage

broerjuang:
## Why

Today the only representation of a sign-in step is HTML. `distill_post_loop` returns a page rendered by `html_renderer.render_form`, so a client that can't embed our page has to scrape it back.

That is exactly what happens: `demos/packages/connector-sdk/src/server/signin-form-schema.ts` re-parses the rendered dpage with cheerio to rebuild a `SigninFormSchema`. Deriving that server-side, from the DOM we already have, removes three defects caused by the round trip:

1. **Completion is detected by an English sentence** — `data.includes('Finished! You can close this window now')`, in two places (`connector/src/core/connector.ts:273`, `connector-sdk/src/server/router.ts:148`).
2. **Error taxonomy is regex-over-HTML** — a 10-rule keyword table (`connector/src/core/mcp-client.ts:265-328`) matching values that began life as a structured `rb-error` attribute.
3. **Grouped radios are unrepresentable** because `SigninFormField` is flat. This is the specific reason the iframe can't be deleted: the OTP-delivery screen (`patterns/nordstrom-mfa-choice.html`) can't be rendered natively.

## What

A JSON sibling of the existing route, in the same router:

```
GET  /dpage/{id}          -> HTML   (unchanged)
POST /dpage/{id}          -> HTML   (unchanged)
GET  /dpage/{id}/json     -> SigninState
POST /dpage/{id}/json     -> SigninState
```

- **`getgather/mcp/form_schema.py`** (new) — `parse_blocks()` projects a distilled document into ordered, typed blocks. Rules derived from the 463 files in `getgather/mcp/patterns/`: hidden-subtree skipping, radio grouping by `name`, and the ad-hoc `for=`-on-non-label-tags association `nordstrom-mfa-choice.html` uses (case-insensitively, since it pairs `id="EMAIL"` with `for="email"`).
- **`distill_post_loop` now returns a `LoopOutcome`** instead of an `HTMLResponse`, with `render_outcome()` and `signin_state()` as the two projections. Order is preserved because it is load-bearing: the prose before a radio group is what explains the options.
- Timeout returns **200 with `status: "TIMEOUT"`** on the JSON path — a JSON client can't tell our 503 apart from a proxy killing a long-poll. The HTML path still raises 503, unchanged.
- Reads use a short poll (`DEFAULT_DPAGE_READ_POLL_TIMEOUT = 5`); a GET submits nothing and re-distilling is idempotent, so it shouldn't hold a request slot for the full submit window (`fly.toml` sets `soft_limit = 20`).
- `button` is now a separate parameter rather than a reserved key inside `fields`, so a site input named `button` can't be mistaken for a branch choice — `google-signin-choose-method.html` declares `<button name="button">`, `<input name="none">` **and** `<button name="submit">`.

## Verification

- `make check` clean (ruff, prettier, yamlfix, ty, pyright strict, pattern check).
- 31 new tests; `parse_blocks` covered against real pattern files, plus the first route-level coverage dpage has had.
- **HTML output proven byte-identical**: all 463 patterns rendered through both the pre-refactor expressions and `render_outcome`, for `need_input` and `finished` with and without an error code — 0 differences.
- Driven end to end against a real remote browser and a live Amazon sign-in: the read returns the email step, and submitting returns `{"status":"ERROR","error_code":"email_not_found"}` — the fact that today's consumer recovers by substring-matching HTML.

## Notes

- `AGENTS.md:79` claims `browser_id` derives from an auth user `{sub}-{auth_provider}`; it does not — it's `mcp_session_id or get_host_id()`. Not changed here.
- The new routes require an explicit `signin_id`, so they never reach the `get_host_id()` fallback (`hostname + "-noauth"`), which is the one genuine horizontal-scaling hazard on the MCP tool path.

### DIFF
--- a/getgather/mcp/dpage.py
+++ b/getgather/mcp/dpage.py
@@ -5,7 +5,7 @@ import re
 import socket
 import urllib.parse
 from dataclasses import dataclass
-from typing import Any, Self
+from typing import Any, Literal, Self
 
 import zendriver as zd
 from bs4 import BeautifulSoup, Tag
@@ -30,6 +30,7 @@ from getgather.browser import (
     zen_navigate_with_retry,
 )
 from getgather.config import settings
+from getgather.mcp.form_schema import SigninState, SigninSubmission, parse_blocks
 from getgather.mcp.html_renderer import DEFAULT_TITLE, render_form
 from getgather.zen_distill import (
     Match,
@@ -52,6 +53,12 @@ router = APIRouter(prefix="/dpage", tags=["dpage"])
 
 # Max seconds for the distillation polling loop in zen_post_dpage (per HTTP request).
 DEFAULT_DPAGE_POST_POLL_TIMEOUT = 60
+
+# Reading the current step submits nothing and re-distilling is idempotent, so a read can poll
+# briefly and let the client retry instead of occupying a request slot for a full minute. Each
+# in-flight request holds one of the machine's connection slots (fly.toml sets soft_limit = 20),
+# which is the real ceiling on concurrent sign-ins.
+DEFAULT_DPAGE_READ_POLL_TIMEOUT = 5
 
 FRIENDLY_CHARS: str = "23456789abcdefghijkmnpqrstuvwxyz"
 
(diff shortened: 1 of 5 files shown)


faisalive:
I think we can apply this changes in backstage instead of here @broerjuang ?

faisalive:
Can we close this @broerjuang ?

broerjuang:
Yes @faisalive thanks for poking me

PR #1430, opened 2026-08-05
feat(amazon): support deployed mock origin

broerjuang:
## Summary

- add an opt-in `AMAZON_BASE_URL` setting for the deployed mock Amazon app from corelens-engineering/demos#1244
- route every Amazon US page and pagination API through the configured origin while leaving Amazon Canada unchanged
- add unit coverage and an opt-in MCP test for all 11 Amazon US tools
- document deployed-mock setup and credentials

## Testing

- `uv run ruff check getgather/config.py getgather/mcp/amazon.py tests/test_amazon.py tests/mcp/test_amazon.py`
- `uv run ruff format --check getgather/config.py getgather/mcp/amazon.py tests/test_amazon.py tests/mcp/test_amazon.py`
- `uv run pytest tests/test_amazon.py tests/mcp/test_amazon.py -q` (3 passed, deployed mock test skipped while unset)
- `uv run pyright getgather/config.py getgather/mcp/amazon.py tests/test_amazon.py tests/mcp/test_amazon.py`
- `npx --yes prettier --check README.md`

## Dependency

The opt-in end-to-end test requires a deployed instance of the mock Amazon app introduced by corelens-engineering/demos#1244 and `AMAZON_BASE_URL` set before Remote Browser starts.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -3,6 +3,9 @@ ENVIRONMENT=local
 PORT=23456
 DATA_DIR=
 
+# Optional deployed Amazon US mock origin. Leave empty to use https://www.amazon.com.
+AMAZON_BASE_URL=
+
 # Local browser backend: "podman" (local containers, default) or "daytona" (on-demand sandboxes).
 # Ignored when CHROMEFLEET_URL is set (requests are proxied to the external Chrome Fleet).
 BROWSER_BACKEND=podman
--- a/README.md
+++ b/README.md
@@ -73,6 +73,28 @@ Follow the MCP install [guide](https://code.visualstudio.com/docs/copilot/chat/m
 
 </details>
 
+### Local mock Amazon
+
+The Amazon US connector can target a deployed instance of the deterministic `apps/mock-amazon`
+app from the [`corelens-engineering/demos` repository](https://github.com/corelens-engineering/demos)
+instead of the real site. Set its browser-reachable origin before starting Remote Browser:
+
+```bash
+AMAZON_BASE_URL=https://your-deployed-mock-amazon.example.com
+```
+
+The override applies to Amazon US only, including its Prime Video pagination APIs. Amazon Canada
+and an unset configuration continue to use the real sites.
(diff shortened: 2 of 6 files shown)


.env.template
```diff
@@ -3,6 +3,9 @@ ENVIRONMENT=local
 PORT=23456
 DATA_DIR=
 
+# Optional deployed Amazon US mock origin. Leave empty to use https://www.amazon.com.
```
ariya, line 6: I don't this needs to be documented here, since it's for internal testing purpose only.

README.md
```diff
@@ -73,6 +73,28 @@ Follow the MCP install [guide](https://code.visualstudio.com/docs/copilot/chat/m
 
 </details>
 
+### Local mock Amazon
```
ariya, line 76: This is not relevant for the user (e.g. who wants to self-host Remote Browser), so perhaps move it to a separate `development.md` or something like that.

README.md
```diff
@@ -73,6 +73,28 @@ Follow the MCP install [guide](https://code.visualstudio.com/docs/copilot/chat/m
...
+
+The Amazon US connector can target a deployed instance of the deterministic `apps/mock-amazon`
+app from the [`corelens-engineering/demos` repository](https://github.com/corelens-engineering/demos)
```
ariya, line 79: Do not reveal private repo.

ariya:
@broerjuang Is it possible to do this via runtime choice instead of deployment flag? An example: an optional parameter in the Amazon MCP tools to supply the target domain.

README.md
broerjuang: sorry, will do

broerjuang:
> is it possible to do this via runtime choice instead of deployment flag? An example: an optional parameter in the Amazon MCP tools to supply the target domain.

We may not need this one. I think the runtime is a better alternative than a deployment flag. I'll implement it on a different PR and close this one.

PR #1431, opened 2026-08-06
feat(amazon): add optional base_url to the US MCP tools

broerjuang:
Supersedes #1430. That PR pointed the Amazon US connector at a stand-in site with a deploy-time `AMAZON_BASE_URL` env var; @ariya asked for a runtime choice instead:

> Is it possible to do this via runtime choice instead of deployment flag? An example: an optional parameter in the Amazon MCP tools to supply the target domain.

## What changed

Each of the 11 Amazon US tools now takes an optional `base_url` — a full origin used instead of `https://www.amazon.com`. The choice is made per call. **There is no environment variable and no deployment setting anywhere**: omit `base_url` and the tools hit real Amazon exactly as before. Amazon CA is unchanged.

`base_url` is a full origin rather than a bare hostname so an `http://` site works.

`AmazonCountry.domain` becomes a `base_url` field — the old derived property was its only reader — plus `with_base_url()`, which rebases every absolute URL onto a new origin. The video URLs are rebased individually rather than derived from the base, since Amazon CA serves them from a different origin than its storefront.

## Testing

- `tests/test_amazon.py` — unit tests for the rebasing, including a reflection guard that fails if a future URL field is added and not rebased. No server, no configuration; runs in `make test`.
- `tests/mcp/test_amazon_alt_origin.py` — drives all 11 US tools through an alternate origin, gated on `--amazon-base-url`. Skips without the flag, so `-m mcp` is unaffected.

`make check` clean; `make test` 143 passed / 12 skipped. Verified by hand that the override applies, the default path is unchanged, CA is unaffected, and an invalid origin surfaces a legible `ToolError`.

### DIFF
--- a/getgather/mcp/amazon.py
+++ b/getgather/mcp/amazon.py
@@ -1,12 +1,14 @@
 import asyncio
 import json
 import os
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from datetime import datetime
-from typing import Any, cast
+from typing import Annotated, Any, cast
+from urllib.parse import urlsplit, urlunsplit
 
 import zendriver as zd
 from loguru import logger
+from pydantic import Field
 
 from getgather.browser import get_url, page_query_selector, zen_navigate_with_retry
 from getgather.mcp.dpage import (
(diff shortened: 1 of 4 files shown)


faisalive:
Do we still need this @broerjuang ?

PR #1435, opened 2026-08-10
DO NOT MERGE — feat(amazon): repoint Amazon at the amazon-mock fixture (deploy branch)

faisalburhanudin:
> **Do not merge this branch. Deploy from it.** It repoints Amazon at a mock fixture; on `main` that would send every production caller at `amazon-mock.dataportrait.app`. `main` merges *in* to pick up upstream changes, never the reverse. Open for review and history, not for merge.

The remotebrowser half of a pair. The other half is the branch of the same name in **corelens-engineering/demos** (PR #1275), which carries the sidecar deployment — `Dockerfile.sidecar`, `start-sidecar.sh`, `sync-remotebrowser.sh` and three Fly configs for `tap-connect-mock-{flyfleet,browserbase,daytona}`.

That repo's `sync-remotebrowser.sh` vendors `getgather` out of a local checkout of *this* repo into its build context, so **whichever branch this repo is on is what gets deployed.** The two branches are only meaningfully deployable together — which is why this one exists rather than the demos side carrying a half-repointing that silently hits real Amazon.

## The change

65 files, essentially all host substitution:

| Files | Change |
| --- | --- |
| `getgather/mcp/amazon.py` | `AMAZON_US.domain` and its five absolute `*_url` fields → `amazon-mock.dataportrait.app` |
| 64 × `getgather/mcp/patterns/*.html` | `rb-domain="amazon"` → `rb-domain="amazon-mock.dataportrait.app"` |

Two edits are not pure substitution and are where review effort is worth spending:

- **`base_url` no longer hard-codes `www.`** — it was `f"https://www.{self.domain}"`, and `domain` is now the full host so a mock host without a `www.` works unchanged. `AMAZON_CA` compensates by carrying `www.amazon.ca` as its domain, so its behaviour is byte-identical to before. Worth confirming that reading.
- **The watch-history `Referer` now interpolates `{country.watch_history_url}`** instead of a literal `https://www.amazon.com/...`. This is a real fix independent of the mock — `_get_watch_history_with_pagination` is country-parameterised, so an `AMAZON_CA` call was sending a `.com` referer. It is the one thing here that would be worth landing on `main` on its own merits.

## What was left on `mock`

One commit, `869e5a9` — five files of loopback-CDP leak fixes. Excluded because they are browser-lifecycle product changes, not mock configuration, and want their own review. `MOCK-OVERRIDES.md` inventories them individually with ship dates and known remainders.

**They matter operationally.** Without them a sidecar deployed from this branch leaks roughly **300 sockets and ~284 MB per run** on the `browserbase` and `daytona` backends, reaching ~600 MB on a 1 GB VM. Past that, memory pressure kills `tailscaled`, the tailnet database goes unreachable, `/health` 500s, Fly deroutes the machine, and every test against it fails *while the machine simultaneously goes dark in telemetry* — which is how it hides. Deployments setting `CHROMEFLEET_URL` (the flyfleet tenant) do not take the leaking path.

So: deployable, but the browserbase and daytona tenants will wedge on a timescale of hours until those fixes land somewhere.

## Review notes

- The 64 pattern files are one mechanical substitution each; `getgather/mcp/amazon.py` and `MOCK-OVERRIDES.md` are the only files worth reading closely.
- Those 65 files are byte-identical to `mock`.
- Evidence for the leak claims lives in the demos repo on `mock`, at `apps/tap-connect/DEBUG-mock-tests.md`.
- Not deployed from this branch; the measurements cited were taken against `mock`.

### DIFF
--- a/MOCK-OVERRIDES.md
+++ b/MOCK-OVERRIDES.md
@@ -1,0 +1,88 @@
+# The amazon-mock overrides, and what this branch leaves behind
+
+`feat/tap-connect-sidecar` carries **the amazon-mock repointing and nothing
+else**, extracted from the `mock` branch.
+
+It is the remotebrowser half of a pair. The other half is the branch of the same
+name in **corelens-engineering/demos**, which carries the sidecar deployment —
+`Dockerfile.sidecar`, `start-sidecar.sh`, `sync-remotebrowser.sh` and three Fly
+configs for `tap-connect-mock-{flyfleet,browserbase,daytona}`. That repo's
+`sync-remotebrowser.sh` vendors `getgather` out of *this* checkout into its build
+context, so **whichever branch this repo is on is what gets deployed.** The two
+branches are meaningfully deployable only together; see `apps/tap-connect/SIDECAR.md`
+and `SIDECAR-GAPS.md` over there.
+
+> **Do not merge this branch.** It repoints Amazon at a mock fixture. On `main`
+> that would send every production caller at `amazon-mock.dataportrait.app`.
+> `main` merges *in* to pick up upstream changes, never the reverse.
+
+Extracted from `mock` at `869e5a9`, against `main` at `181421c`.
+
+## What the repointing is
+
+65 files, every one of them a host substitution:
+
+| Files | Change |
+| --- | --- |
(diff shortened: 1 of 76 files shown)


ariya:
@faisalburhanudin Can we close this for now?

PR #1440, opened 2026-08-12
perf(distill): cache parsed distillation patterns

faisalburhanudin:
## What

Adds `@lru_cache(maxsize=8)` to `load_distillation_patterns`, which re-read and re-parsed all **470** pattern files on every call. Every MCP tool call went through it.

## Measured

On the `tap-connect-mock-flyfleet` sidecar (`shared-cpu-2x`, Fly `sjc`), both arms run back-to-back in one process against the same files. Arm A calls `.__wrapped__` — the undecorated function — so it is the genuine pre-cache path. `process_time` is used rather than wall clock because it is unaffected by host CPU steal.

| arm | CPU per call |
|---|---|
| uncached | 0.2174 / 0.2170 / 0.2162 / 0.2192 s |
| `lru_cache` | 0.0000 s (4 hits, 1 miss) |

`py-spy` at 200 Hz, 1570 samples inside the function:

| share | where |
|---|---|
| 55.9% | bs4 tree build |
| 29.9% | `html.parser` tokenize |
| 9.0% | `glob` + file read |

So ~86% of the cost is BeautifulSoup parsing, and the cache eliminates it entirely after the first call.

## Why it is safe

- `Pattern.pattern` is read in exactly one place repo-wide — `distill()` — and it `deepcopy`s the tree before mutating it.
- Nothing mutates the returned list in place (no `append`/`sort`/`pop` outside the constructor loop), which matters because `lru_cache` shares the list object too, not just the trees.
- `path` is a `str`, so it is hashable.
- `maxsize=8` covers the distinct globs in use: `mcp/patterns/**`, the dpage patterns, `main.py`'s single `ip-fly.html`, and the test fixtures.

### DIFF
--- a/getgather/patterns_api_router.py
+++ b/getgather/patterns_api_router.py
@@ -6,6 +6,8 @@ from fastapi.responses import JSONResponse, PlainTextResponse
 from loguru import logger
 from pydantic import BaseModel
 
+from getgather.zen_distill import load_distillation_patterns
+
 router = APIRouter()
 
 PATTERNS_DIR = os.path.join(os.path.dirname(__file__), "mcp", "patterns")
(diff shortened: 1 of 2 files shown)


ariya:
The cache needs to be cleared when there is a pattern upload.

faisalburhanudin:
> The cache needs to be cleared when there is a pattern upload.

Got it mas Ariya

ariya (approved):

PR #1443, opened 2026-08-13
fix(browser): add 30s timeout to get_remote_browser chromefleet check

faisalive:
## Problem
`get_remote_browser` calls `call_chromefleet_api("GET", browser_id)` with no timeout override, so it falls back to the default 120s x 3 retries. Traced a live incident (`/dpage/{id}` POST) where this stalled the request until the caller's own undici `headersTimeout` (300s) fired first, surfacing as a bare `Headers Timeout Error` with no useful error attached.

Sibling `get_remote_browser_cdp_url` already passes `timeout=20.0` — this one was missed.

## Fix
Pass `timeout=30.0` explicitly, bounding the worst case well under the caller's 300s budget.

## Test plan
- [x] `make check-backend-format`
- [x] `make typecheck`

### DIFF
--- a/getgather/browser.py
+++ b/getgather/browser.py
@@ -221,7 +221,7 @@ async def get_remote_browser_cdp_url(browser_id: str) -> str:
 async def get_remote_browser(browser_id: str) -> zd.Browser | None:
     logger.debug(f"Finding the ChromeFleet browser: {browser_id}")
     try:
-        await call_chromefleet_api("GET", browser_id)
+        await call_chromefleet_api("GET", browser_id, timeout=30.0)
     except Exception:
         return None
 


getgather/browser.py
```diff
@@ -221,7 +221,7 @@ async def get_remote_browser_cdp_url(browser_id: str) -> str:
 async def get_remote_browser(browser_id: str) -> zd.Browser | None:
     logger.debug(f"Finding the ChromeFleet browser: {browser_id}")
     try:
-        await call_chromefleet_api("GET", browser_id)
+        await call_chromefleet_api("GET", browser_id, timeout=30.0)
```
broerjuang, line 224: is 30s enough to decide?

faisalive: Yaps, it's for gate so we dont have infinite loading/timeout.

faisalburhanudin (approved):

PR #1451, opened 2026-08-14
feat(dpage): CloakBrowser humanize and reload-before-actions

faisalive:
## Summary
- Add CloakBrowser humanize adapters (`cloak_human.py`) and per-pattern `rb-humanize` typing/clicks
- Support `rb-reload-before-actions`, settle/idle/submit delay attrs in the dpage loop
- Depend on `cloakbrowser>=0.5.0` via the daytona extra

Stacked under: CVS bot-detection will target this branch.

## Test plan
- [ ] `uv run pytest tests/test_cloak_human.py tests/mcp/test_dpage_reload.py`
- [ ] `make check`
- [ ] Confirm no CVS pattern files are in this PR


broerjuang:
For this, I think we should use custom plugins first

PR #1453, opened 2026-08-14
fix(cdp): namespace targetIds in Target.getTargets responses

faisalive:
## Problem

Traced a 13.5s stall inside a `/dpage` sign-in flow. The `/devtools/page/<id>` websocket handler spent it scanning the entire live fleet — 1 `AsyncDaytona.list`, 16 `AsyncDaytona.get`, ~12 `/json/version` probes — just to map a `page_id` back to its `browser_id`. The target sandbox matched last.

Root cause is in `patch_cdp_target`. It matched a handful of known message shapes:

| CDP message | field carrying the id | patched? |
|---|---|---|
| `Target.targetCreated` | `params.targetInfo.targetId` | ✅ |
| generic result | `result.targetId` | ✅ |
| **`Target.getTargets`** | **`result.targetInfos[]`** (a list) | ❌ |
| `Target.attachedToTarget` | `params.targetInfo.targetId` | ❌ |
| `Target.targetInfoChanged` | `params.targetInfo.targetId` | ❌ |

zendriver's `update_targets()` calls `Target.getTargets` and builds each tab's websocket URL directly from that list (`zendriver/core/browser.py:568`):

```python
f"ws://{host}:{port}/devtools/page/{t.target_id}"
```

An unpatched id there means a raw URL with no `browser_id`, so `cdp_devtools_websocket_proxy` misses the `@` branch and falls through to `find_browser_id`, which probed every live browser **in series** at ~0.9s each.

That scales badly:

| live sandboxes | before | after |
|---|---|---|
| 16 (observed) | 13.5s | 0 |
| 100 | ~90s | 0 |
| 500 | ~7.5min | 0 |

## Changes

**Namespace every `targetId` (the actual fix).** Replaced the shape matching with a recursive walk over the decoded message, split by direction:

- `patch_cdp_target_inbound` (browser → client) — prepends `browser_id@` to every `targetId` at any depth, **including inside lists**
- `patch_cdp_target_outbound` (client → browser) — strips on any method, not the three-method allowlist that missed `Target.activateTarget` and `Browser.getWindowForTarget`
- `prepend_browser_id_to_target_id` is now idempotent, so a re-patched id can't double-prefix

Walking the tree rather than enumerating shapes is the point: an allowlist silently misses whatever CDP message nobody thought of, and the failure mode is a latency cliff proportional to fleet size rather than an error.

**Bound the fallback (safety net).** `find_browser_id` now probes in parallel with a concurrency cap, first hit wins, remaining tasks cancelled. Log level `debug` → `warning`, since reaching it now means a routing regression upstream rather than a normal path.

## Testing

6 new tests in `tests/test_browser_backend.py`: `targetInfos[]` list, `Target.attachedToTarget`, `Target.targetCreated`, idempotency, outbound strip across five methods, and pass-through for messages with no target ids.

`make check` clean. Unit suite **140 passed, 0 failed** (baseline on `main`: 134 passed).

## Not covered

`openerId` and `browserContextId` are also target-ish ids and stay raw — unchanged from current behaviour. Nothing routes on them today, but worth knowing they're inconsistent with `targetId` now.


broerjuang:
The recursive target-ID rewrite looks too broad and can silently corrupt application data carried through CDP.

`_rewrite_target_ids` rewrites every JSON property named `targetId` at any depth. That includes opaque values transported by methods such as `Runtime.evaluate` and `Runtime.callFunctionOn`, rather than only fields owned by the CDP Target protocol.

I reproduced both directions on this branch:

```json
// Runtime.evaluate response
{"value":{"targetId":"order-123"}}
// becomes
{"value":{"targetId":"BROWSER@order-123"}}
```

```json
// Runtime.callFunctionOn argument
{"value":{"targetId":"user@domain"}}
// becomes
{"value":{"targetId":"domain"}}
```

This service frequently returns structured website data through `page.evaluate`, so a site object containing a `targetId` property could be modified without an error.

Could we restrict rewriting to protocol-owned locations, including `result.targetInfos[*].targetId`, `params.targetInfo.targetId`, and `params.targetId` for the relevant Target/Browser commands? Please also add regression tests proving that `targetId` inside Runtime values and arguments passes through unchanged.

PR #1458, opened 2026-08-18
feat(browser): race providers behind neutral CDP proxy

broerjuang:
## What this does

Adds an opt-in best-of-providers experiment for server-assigned browser creation.

When `BROWSER_PROVIDER_RACE=true`, `POST /api/v1/browsers` starts one candidate on every configured remote provider:

- Fly / external Chrome Fleet (`CHROMEFLEET_URL`)
- Daytona (`DAYTONA_API_KEY` + `DAYTONA_SNAPSHOT`)
- Browserbase (`BROWSERBASE_API_KEY`)

A candidate wins only after its CDP connection successfully answers `Target.getTargets`. The first CDP-ready browser is returned; remaining candidates finish in the background and are cleaned up asynchronously.

## Compatibility

The create response keeps the previous contract:

```json
{
  "browser_id": "Bdabc1234",
  "status": "created"
}
```

Clients continue constructing the proxied CDP endpoint as `/cdp/{browser_id}`. Direct provider CDP URLs, signed credentials, and provider-assigned Browserbase session IDs are never returned.

- The experiment is disabled by default.
- `POST /api/v1/browsers/{browser_id}` is unchanged.
- GET, DELETE, CDP, DevTools, and live-view requests continue using the returned `browser_id`.
- CDP target-ID namespacing remains specific to the winning provider.

## Routing without shared state

Server-assigned IDs remain nine characters. Their second character records the winning provider:

- `Bf.......` — Fly / external Chrome Fleet
- `Bd.......` — Daytona
- `Bb.......` — Browserbase

This lets any GetGather replica route later requests without Redis or an in-memory winner registry. The accepted tradeoff is that the winning provider is visible from the ID prefix.

Fly and Daytona use the routed ID as their native browser ID. Browserbase assigns its own session ID, so creation attaches the routed `Bb...` ID as user metadata. After a restart or replica hop, GetGather queries that metadata and fetches the current Browserbase connection URL.

## Enabling the experiment

```env
BROWSER_PROVIDER_RACE=true
CHROMEFLEET_URL=...
DAYTONA_API_KEY=...
DAYTONA_SNAPSHOT=...
BROWSERBASE_API_KEY=...
```

Only configured providers participate, and at least two are required. The deployment image must include the optional Daytona SDK when Daytona participates.

## Validation

- `make check`
- 157 unit tests passed
- Live smoke tests covered three-provider racing, two-provider racing, fallback when Browserbase returns 401, proxied CDP commands, and browser deletion.
- Automated tests cover fastest-ready selection, provider failures, loser cleanup, routed-ID recovery across process boundaries, Browserbase metadata recovery, log redaction, and the unchanged create response contract.

## Remaining follow-up

1. Collect canary latency, winner distribution, provider failure, and cleanup metrics.
2. Add durable reconciliation for losing candidates that survive a process crash or exhaust cleanup retries.
3. Add cost and concurrency guardrails before broad rollout.
4. Decide later whether provider racing should cover named creation and MCP/dpage flows; this PR intentionally limits it to server-assigned browser creation.

### DIFF
--- a/getgather/browsers/browserbase_browsers.py
+++ b/getgather/browsers/browserbase_browsers.py
@@ -1,7 +1,7 @@
 import asyncio
 import json
 import os
-from typing import Any
+from typing import Any, cast
 
 import httpx
 import websockets
@@ -170,8 +170,9 @@ class BrowserbaseBackend:
 
     Only `create_browser` is implemented: it POSTs to the Browserbase `/v1/sessions`
     endpoint with the `x-bb-api-key` header (read from the `BROWSERBASE_API_KEY` env var)
-    and stores the returned `id -> connectUrl` mapping in an in-memory hashmap so the
-    CDP proxy in `router.py` can later route `/cdp/<id>` traffic to the right connectUrl.
+    and stores the returned `id -> connectUrl` mapping locally. It also tags sessions with the
+    routed GetGather browser ID, allowing another process to recover the assigned ID and
+    connectUrl from Browserbase after a restart or replica hop.
 
     The Browserbase-assigned session `id` is returned as the `browser_id`, ignoring the
     caller-supplied id (Browserbase creates its own). The router's `POST /api/v1/browsers`
@@ -183,6 +184,7 @@ class BrowserbaseBackend:
     def __init__(self) -> None:
         # browser_id -> connectUrl (wss://connect.usw2.browserbase.com/?signingKey=...)
         self._sessions: dict[str, str] = {}
+        self._logical_sessions: dict[str, str] = {}
 
     async def shutdown(self) -> None:
         return None
@@ -203,7 +205,10 @@ class BrowserbaseBackend:
         # keepAlive keeps the session running between CDP connections / after disconnects.
         # Without it, Browserbase ends the session the moment the first WS drops, and any
         # subsequent connect to the same signingKey returns HTTP 410 Gone.
-        body: dict[str, Any] = {"keepAlive": True}
+        body: dict[str, Any] = {
+            "keepAlive": True,
+            "userMetadata": {"getgatherBrowserId": browser_id},
+        }
         proxy_config = await get_proxy_config(origin_ip, target_domain, settings)
         if proxy_config:
             proxy_url = proxy_config.get_proxy_url(browser_id)
@@ -219,6 +224,7 @@ class BrowserbaseBackend:
         bb_id = str(data["id"])
         connect_url = str(data["connectUrl"])
         self._sessions[bb_id] = connect_url
+        self._logical_sessions[browser_id] = bb_id
         # connectUrl embeds a signing key and is a bearer secret; never log it.
         logger.info(f"Browserbase session created: id={bb_id}")
         await self._wait_until_cdp_ready(connect_url, bb_id)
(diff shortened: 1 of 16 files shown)


broerjuang:
Live testing found a post-delete routing issue: after a successful DELETE, the in-memory winner route was removed, so a subsequent GET could fall through to the default Fly backend and return 500. Commit ff4a1c3 retains the provider route as a tombstone, keeps CDP auto-start affinity on the original winner, excludes deleted routes from listings, and adds a create -> delete -> GET regression test. Current unit result: 138 passed, 12 skipped, 30 deselected; pyright and ty pass.

broerjuang:
Update: redacted provider CDP URLs from logs and added simple timing logs for provider readiness, winner selection, failures, and cleanup. Client behavior is unchanged. Tests and type checks pass.

getgather/browsers/provider_race.py
```diff
@@ -0,0 +1,350 @@
...
+            provider_browser_id=winner.provider_browser_id,
+        )
+        logger.info(
+            f"Provider-race winner for {browser_id}: provider={winner.provider} "
+            f"create_ms={winner.create_seconds * 1000:.0f} "
+            f"cdp_ready_ms={winner.cdp_ready_seconds * 1000:.0f} "
+            f"race_ms={(time.monotonic() - race_started) * 1000:.0f}"
+        )
+
+        cleanup_task = asyncio.create_task(
```
broerjuang, line 95: We intentionally leave the pending provider tasks running after selecting a winner. Cancellation does not guarantee that the remote allocation stops; waiting for each task to settle gives us the provider-assigned ID needed to delete the losing session.

getgather/browsers/provider_race.py
```diff
@@ -0,0 +1,350 @@
...
+        try:
+            result = await provider_backend.create_browser(
+                public_browser_id, origin_ip, target_domain, browser_type
+            )
+            created_at = time.monotonic()
+            result_browser_id = result.get("browser_id")
+            if isinstance(result_browser_id, str):
+                provider_browser_id = result_browser_id
+            phase = "cdp_ready"
+            await self._wait_until_cdp_ready(provider_backend, provider_browser_id)
```
broerjuang: A provider only wins after this CDP readiness probe succeeds. Create APIs can return before Chrome accepts commands, so racing on create completion alone would occasionally hand the client an unusable browser.

getgather/browsers/provider_race.py
```diff
@@ -0,0 +1,350 @@
...
+        self, browser_id: str, origin_ip: str | None, target_domain: str | None
+    ) -> dict[str, Any]:
+        provider_backend, provider_browser_id = self._route(browser_id)
+        await provider_backend.get_browser(provider_browser_id, origin_ip, target_domain)
+        return {"browser_id": browser_id, "status": "created"}
+
+    async def delete_browser(self, browser_id: str) -> dict[str, Any]:
+        provider_backend, provider_browser_id = self._route(browser_id)
+        await provider_backend.delete_browser(provider_browser_id)
+        # Keep the route as a tombstone. Falling through to the default backend after deletion can
```
broerjuang: The route deliberately survives deletion as a tombstone. Removing it caused later GET and CDP requests to fall through to the default backend, producing a 500 or recreating the same public ID on a different provider.

getgather/browsers/browserbase_browsers.py
```diff
@@ -367,7 +369,7 @@ async def delete_browser(self, browser_id: str) -> dict[str, Any]:
                     response.raise_for_status()
             logger.info(f"Browserbase session released: id={browser_id}")
         except Exception as e:
-            logger.warning(f"Browserbase release failed for {browser_id}: {type(e).__name__}: {e}")
+            logger.warning(f"Browserbase release failed for {browser_id}: error={type(e).__name__}")
```
broerjuang, line 458: The forced invalid-key fallback run showed that the httpx exception text includes the full provider API URL. Logging only the exception class here preserves the failure signal without leaking the provider endpoint; the regression test verifies the URL and response detail stay out of logs.

getgather/browsers/route_id.py
```diff
@@ -0,0 +1,29 @@
+from dataclasses import dataclass
+
+PROVIDER_CODES = {"fly": "f", "daytona": "d", "browserbase": "b"}
```
broerjuang, line 3: These three prefixes are deliberately reserved for server-assigned provider-race IDs. Keeping the ID at nine characters preserves the previous shape; the accepted tradeoff is that the winner is now visible from the second character.

getgather/browsers/browserbase_browsers.py
```diff
@@ -316,9 +323,81 @@ async def get_cdp_websocket_remote_url(self, browser_id: str) -> str | None:
...
 
-    def cdp_targets_need_namespacing(self) -> bool:
+    async def resolve_session(self, logical_browser_id: str) -> str | None:
+        """Recover a Browserbase session after a GetGather restart or replica hop."""
+        cached_id = self._logical_sessions.get(logical_browser_id)
+        if cached_id is not None and cached_id in self._sessions:
+            return cached_id
+
+        headers = {"x-bb-api-key": _api_key()}
+        query = f"user_metadata['getgatherBrowserId']:'{logical_browser_id}'"
```
broerjuang, line 339: Browserbase is the only provider that cannot use the routed ID as its native session ID. We attach the `Bb...` value as user metadata during creation, query it here after a cache miss, and then fetch the current connectUrl without persisting that signed URL in GetGather.

bin-ario:
good idea to make the browser infra more robust
code-wise, `provider_race.py` can be simpler if you move browser backend specific things (e.g., check cdp readiness) to individual backend module, because they share the same protocol

PR #1464, opened 2026-08-23
DON'T MERGED feat(daytona): log-only probe for what keeps sandboxes out of auto-stop

faisalburhanudin:
## Why

Daytona billing data (200 sandboxes, \$181.39) shows **58 sandboxes lived >1h and cost \$66.86**, and among those the average *stopped* time is **0.06h** — `Billed Running (h)` ≈ `Lifetime (h)`. `auto_delete_interval` works fine (142 auto-deletes); what doesn't is `auto_stop_interval`: something resets the 15-minute idle clock continuously, for up to 46 hours.

This adds a probe to find out which function, before changing any behavior.

## What it records

`getgather/browsers/daytona_probe.py`:

- **Sandbox touches** — every call this process makes into a sandbox: `process.exec` (history read, proxy config, IP check), signed preview URLs, `start`, `create`, `delete`, `api.get` — each attributed to the nearest app-level caller frame (`dpage.py:211 poll`, etc).
- **Long-lived connections through the preview proxy** — the CDP relay in `websocket_proxy`, and zendriver `Browser` handles, which nothing closes unless a caller reaches `terminate_remote_browser`. An open socket is sandbox activity even when no API call is made, so this is the candidate an exec-only probe would miss.

Every 5 minutes the existing background loop emits one report per browser: touch counts by op, top callers, max gap between touches, open connections and their ages, and a **keepalive verdict** (no gap ever exceeded the idle window, or a connection outlived it).

## Log-only

- No Daytona API calls of its own; the reporter reads process-local state.
- No lifecycle decision changed; the diff is 36 added lines across 4 files plus the new module.
- Disabled unless `DaytonaBackend` is constructed in the process, so podman/browserbase deployments emit nothing.
- Reports and connection open/close are INFO; per-touch events are DEBUG (aggregates ride in the report).
- Not tracked: the noVNC live view. That iframe connects the viewer's browser straight to Daytona's preview proxy, so this process only sees the URL handout.

## Reading it

In Logfire, filter `attributes->>'probe' = 'daytona'`:
- `daytona.probe.report` — per-browser verdict, `top_callers` names the culprit
- `daytona.probe.conn` — `held_seconds`, `exceeded_auto_stop`
- `daytona.probe.summary` — `keepalive_browsers` per tick

## Checks

`make check` clean (ruff, pyright strict, ty, prettier, yamlfix); `make test` 134 passed, 12 skipped.

### DIFF
--- a/getgather/browser.py
+++ b/getgather/browser.py
@@ -19,6 +19,7 @@ from zendriver.core._contradict import ContraDict
 from zendriver.core.config import Config
 from zendriver.core.connection import Connection, ProtocolException
 
+from getgather.browsers import daytona_probe
 from getgather.browsers.fleet_browsers import build_chromefleet_headers, call_chromefleet_api
 from getgather.config import FRIENDLY_CHARS, settings
 
@@ -164,6 +165,10 @@ async def _create_browser_from_cdp_websocket(
                 f"browser_id={browser_id}"
             )
     instance.id = browser_id  # type: ignore[attr-defined]  # ty: ignore[unresolved-attribute]
+    # Nothing stops this handle unless a caller reaches terminate_remote_browser, and its CDP
+    # websocket stays open through the sandbox's preview proxy meanwhile. Record it so the probe
+    # can report which caller holds one open past Daytona's idle window.
+    daytona_probe.open_connection(browser_id, "zendriver-cdp")
     return instance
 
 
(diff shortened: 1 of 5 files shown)


faisalburhanudin:
Closing: deploying this from local via the fly CLI instead of the CI-on-main path.

PR #1468, opened 2026-08-26
fix: race between browser deletion and browser resurection

faisalburhanudin:
## Why

In very rare cases, we sometimes resurrect a browser that was recently deleted. When this happens, we may end up using a browser that is in the middle of being destroyed. In the case of Daytona, this extends the lifecycle of the deleted sandbox.

## Changes

With this approach, we avoid accidentally choosing an instance that was just destroyed. By maintaining a pool of browsers that were deleted within the last 60 seconds.

### DIFF
--- a/getgather/browsers/backend.py
+++ b/getgather/browsers/backend.py
@@ -12,6 +12,28 @@ from getgather.config import FRIENDLY_CHARS, settings
 # `chromium-abc`. Both local backends derive names and parse ids from this single prefix.
 BROWSER_NAME_PREFIX = "chromium-"
 
+RECENTLY_DELETED_SECONDS = 60.0
+_deleted_at: dict[str, float] = {}
+
+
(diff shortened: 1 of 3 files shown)


getgather/browsers/backend.py
```diff
@@ -12,6 +12,28 @@
 # `chromium-abc`. Both local backends derive names and parse ids from this single prefix.
 BROWSER_NAME_PREFIX = "chromium-"
 
+RECENTLY_DELETED_SECONDS = 60.0
+_deleted_at: dict[str, float] = {}
```
faisalive, line 16: I belive we're using 5 instance for remotebrowser. I belive this wont work for multi-instance app?

faisalburhanudin: You are right about that. Now I'm not sure how to fix it 😆

